PR #1, opened 2017-01-12
Upgrade to Hyper 0.10

bruceadams:
This appears to be required for me to be able to use Hyper 0.10.0.
There is so much here that I do not understand.
I do not know if making this change is appropriate in your library.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hyper-rustls"
-version = "0.2.1"
+version = "0.3.0"
 authors = ["Joseph Birr-Pixton <jpixton@gmail.com>"]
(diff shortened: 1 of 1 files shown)


hannobraun:
I don't know enough about this library to speak with any authority, but I believe there's one mistake in this pull request.

Updating a dependency to a new incompatible version (as you are doing here) can break a library user's build, and must therefore be considered a breaking change.  This means the version bump to `0.2.2` is not sufficient. The new version should be `0.3.0`.

bruceadams:
Ah, yes! You are certainly correct. I've amended my commit. Thanks @hannobraun for catching my versioning mistake and pointing it out!

ctz:
Thanks!

PR #2, opened 2017-01-31
Fix endless loop in read when there is io error in underlying_write.

mkocot:
When this happens we early exit from `underlying_read` so flag for eof
is never set. Because of this all `underlying_*` methods now returns
early. Reading again from session data returns 0 which means exit
conditions from loop are never met.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -71,6 +71,10 @@ impl TlsStream {
     };
   }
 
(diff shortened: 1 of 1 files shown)


ctz:
Sorry about that, and thanks for the fix. 0.3.1 is published

PR #5, opened 2017-03-02
use url crate instead of parsing this

tshepang:


ctz:
These examples are intended to show the minimal changes needed over the original hyper examples of the same name. I've put a bug in to bring these up to date, which fixes this manual parsing in a different way. Thanks for pointing that out!

PR #6, opened 2017-03-17
Adding TlsClient::new_with_ca(&str) method

mikeokner:
Allow users to add a .pem file to the trust store to avoid errors in
situations when servers are using a private CA, clients are behind
a TLS proxy, etc.


ctz:
Sorry for the delay. This seems like a reasonable thing to want. I think I'd prefer to let people pass in a whole `Arc<rustls::ClientConfig>`, then they can do all the file IO and error handling outside `TlsClient`.

ctz:
#28 should address this

PR #12, opened 2017-05-06
Allow getting at underlying rustls::Session

ctz:


alex:
Unfortunately this doesn't work for me, this requires a mutable borrow of the `WrappedStream`, which one doesn't have when using `request.ssl()`: https://hyper.rs/hyper/v0.10.9/hyper/server/request/struct.Request.html#method.ssl

alex:
btw, thanks for jumping on this so quickly!

ctz:
> Unfortunately this doesn't work for me

Oops! Try now.

alex:
Perfect! Works great now, thanks!

PR #19, opened 2017-05-16
WIP: update to use hyper connectors

crawford:
This is still very much a work in progress. I wanted to put it up here to get feedback before I dig too deep. Right now, I'm copying [stream.rs](https://github.com/hyperium/hyper-tls/blob/62e3376736a9d3693c916641e80fdd7e164d97e8/src/stream.rs) and the server example doesn't work. So far, it seems to work nicely. I was able to pull it into [github-rs](https://github.com/mgattozzi/github-rs) and use it in a test program.

Begins to address https://github.com/ctz/hyper-rustls/issues/9.


examples/client.rs
```diff
@@ -1,90 +1,43 @@
...
-}
+use std::io::Write;
+use std::str::FromStr;
 
 fn main() {
-    env_logger::init().unwrap();
-
     let url = match env::args().nth(1) {
-        Some(url) => url,
+        Some(url) => Uri::from_str(&*url).expect("well-formed URI"),
```
lucab: This `&*` sigil seems a bit weird to me, and the "url" name-clash as well. Would a `Some(ref arg_uri)` work here?

crawford: Yeah, `ref` is way cleaner than `&*` or `&[..]`. Thanks.

src/connector.rs
```diff
@@ -0,0 +1,88 @@
...
+use rustls::ClientConfig;
+use std::{fmt, io};
+use std::sync::Arc;
+use stream::MaybeHttpsStream;
+use tokio_core::reactor::Handle;
+use tokio_rustls::ClientConfigExt;
+use tokio_service::Service;
+use webpki_roots;
+
+pub struct HttpsConnector {
```
lucab: Could you please `derive(Clone)` on this?

lucab:
In order to bump this, I went ahead and added the missing pieces (missing derive, server example, deps refreshing) in https://github.com/ctz/hyper-rustls/pull/22.

crawford:
@lucab Ah nice! Let's go with your update.

PR #25, opened 2017-06-17
cargo: update depends and bump to 0.8.0

martell:
- bump webpki-roots, older webpki crates have been removed from crates.
- Fix multiple versions of the ring package because of tokio-rustls and
 rustls. Bump rustls to match the version and specify the minor
 tokio-rustls version to prempt similar problems in future.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,10 +11,10 @@ repository = "https://github.com/ctz/hyper-rustls"
 [dependencies]
 futures = "0.1.13"
 hyper = "0.11"
-rustls = "0.8"
+rustls = "0.9.0"
 tokio-core = "0.1.7"
 tokio-io = "0.1.1"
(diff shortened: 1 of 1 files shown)


martell:
I created https://github.com/ctz/hyper-rustls/pull/25
We will have to wait for that merge to fix hyper-rustls

lucab:
+1 on this, I'm hitting the same deps problem.

martell:
@ctz could we get a merge of the update for `rustls` and `webpki-roots` from this into a 0.6.1 create.
Many projects will not be able to update to hyper 0.11 on such short notice and would be nice if the 0.6.x creates still worked for at least a while longer.

Hmm, In hind sight this probably should have been 0.7.1 and not 0.8.0.
I believe that is why I left the bump out of the actual commit but never updated the message.

PR #31, opened 2017-07-25
Made tlsclient cloneable

SimranGujral:
I am checking Hyper-Rustls on Servo. I have worked on replacing Rust-OpenSSL with Hyper-Rustls.
This requires TLSClient to be cloneable. This PR does that.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -187,6 +187,7 @@ impl NetworkStream for WrappedStream {
   }
 }
 
+#[derive(Clone)]
 pub struct TlsClient {
   pub cfg: Arc<rustls::ClientConfig>
 }


SimranGujral:
We will be working on bumping servo to hyper 0.11 so that we can have backward compatibility with Hyper-Rustls. This is a part of a proof of concept to check the performance of Hyper-Rustls with Servo. If we can have the merge then we can directly point to this branch instead of my local one.

SergioBenitez:
@SimranGujral Feel free to use [SergioBenitez/hyper-sync-rustls](https://github.com/SergioBenitez/hyper-sync-rustls), where `TlsClient` is already cloneable. The library is identical to this one except it targets hyper 0.10.

SimranGujral:
Thanks @SergioBenitez , Also, is there a way to explicitly set ciphers in Rustls client config? I couldn't see one.

lucab:
@SimranGujral have you seen https://docs.rs/rustls/0.9.0/rustls/struct.ClientConfig.html#structfield\.ciphersuites ?

SimranGujral:
Yes, I have. :) I was thinking of working on a function which can do it the way its done in OpenSSL. 
I think I'll try my hand at this first.  Thanks  @lucab  @SergioBenitez :)

SimranGujral:
Closing this. Will use the hyper-async version for now.

PR #46, opened 2018-06-28
examples/server: warn when aborting on connection errors

lucab:
As the example server is used for CI purposes, it is convenient that
it aborts on any client error. However this is not the typical behavior
expected from a server, so call it out explicitly.

Ref: https://github.com/ctz/hyper-rustls/issues/45

### DIFF
--- a/examples/server.rs
+++ b/examples/server.rs
@@ -51,6 +51,7 @@ fn load_private_key(filename: &str) -> rustls::PrivateKey {
 }
 
 fn main() {
+    // First parameter is port number (optional, defaults to 1337)
     let port = match env::args().nth(1) {
         Some(ref p) => p.to_owned(),
         None => "1337".to_owned(),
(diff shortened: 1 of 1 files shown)


lucab:
@azastrael this should make a bit more clear the behavior you experienced. Any feedback from your side is appreciated.

azastrael:
The issue is that it will still exit when an error is received and that it should be explained how to stop it from happening.

For example, here's my way of stopping it from crashing (using the latest version of `tokio`):
```rust
[...]

use hyper::server::conn::Http:

[...]

let tcp = TcpListener::bind(&addr).unwrap();
println!("Starting to serve on https://{}.", addr);
let fut = tcp.incoming().for_each(move |s| {
    tokio::spawn(
        tls_cfg.accept_async(s).and_then(|s| {
            Http::new()
                .serve_connection(s, service_fn(echo)
                .map_err(|e| println!("Error: {:?}", e))
                .wait()
                .unwrap();

            Ok(())
        })
        .map_err(|e| println!("Error: {:?}", e))
    );

    Ok(())
});

tokio::run(fut);
```

lucab:
@azastrael it will always keep exiting for testing purposes, I think. The question is how to tell people copying this to do proper error handling. That's why I added a `map_err` and the comment.

azastrael:
@lucab I don't see how to handle errors using your `map_err`, though: stopping returning the `e` should stop the server from crashing but it will actually just make the compiler throw an error at you (`expected (), found struct std::io::Error`).

Could you explain to me how to use your new example to stop the server from crashing when receiving an error? Thank you and keep up the good work.

lucab:
@azastrael I see. You were looking for something plug-play, but I was just calling out the lack of error catching. I amended this PR so that there is now a single-line-comment change to switch the behavior, but this is likely not the typical way of writing such logic.

azastrael:
@lucab Thank you! :heart:

PR #47, opened 2018-07-15
Add HttpsConnector::new_with_handle

matthauck:


matthauck:
Context: I'm just now trying to upgrade a project to the latest version of hyper/rustls and friends, and trying to deal with API changes. I previously had used the `HttpsConnector::new(threads, handle)` constructor. This PR is an attempt to add that back in the same form that the `HttpConnector` has as well. I noticed that `hyper-tls` also does not have this `new_with_handle` constructor either. 

Is this API being dropped for a reason? 

I suppose it is more convenient to not have to specify a `Handle`, but I'm not entirely sure what that means for async requests, and which Core is then responsible for driving the request.

lucab:
There is a `from(args: (T, ClientConfig))` where you can pass your own `HttpConnector`, would that not work for your usecase?

matthauck:
Ah, yes, that should work just fine.

lucab:
@matthauck I guess this can be closed then?

matthauck:
Yup!

PR #50, opened 2018-08-29
Remove tokio-core and replace examples with tokio

mehcode:


lucab:
@mehcode thanks, look good! I did some cleanups to the examples in https://github.com/ctz/hyper-rustls/pull/49 but it is still not reviewd. Would you mind having a look at that and see if this PR can be easily rebased on that?

examples/client.rs
```diff
@@ -54,8 +54,8 @@ fn main() {
             println!("Body:\n{}", String::from_utf8_lossy(&body));
         });
 
-    let mut core = tokio_core::reactor::Core::new().unwrap();
-    if let Err(err) = core.run(fut) {
+    let mut rt = tokio::runtime::Runtime::new().unwrap();
+    if let Err(err) = rt.block_on(fut) {
```
lucab: Shouldn't this be a `rt.block_on_all` (and the others in this PR too)?

lucab:
@mehcode would you mind rebasing on current master? Also I have a minor question on executor usage.

ctz:
BTW this is merged as-is but I'll keep this PR open to address those questions.

lucab:
Ah dang, I missed that. Then I guess we can close https://github.com/ctz/hyper-rustls/issues/51 as well. I'm still not super familiar with the new tokio runtime, thus keeping the `block_on` vs `block_on_all` question here.

PR #52, opened 2018-11-10
Bumped rustls and tokio-rustls versions

alex:


alex:
nightly-2018-06-11 are failing, but current nightly is fine. These don't look like anything I could have caused.

ctz:
Thanks!

lucab:
@ctz maybe we can turn on dependabot here? https://dependabot.com/rust/

ctz:
I did actually yesterday :)

PR #57, opened 2019-01-03
Add hyper support for ALPN h2

seanmonstar:


src/connector.rs
```diff
@@ -2,7 +2,7 @@ use ct_logs;
 use futures::{Future, Poll};
 use hyper::client::connect::{self, Connect};
 use hyper::client::HttpConnector;
-use rustls::ClientConfig;
+use rustls::{ClientConfig, ClientSession};
```
lucab: I think this was supposed to be `Session` instead.

seanmonstar: Woops, yep! All fixed.

PR #58, opened 2019-01-03
Forward read_buf and write_buf for MaybeHttpsStream

seanmonstar:
This is more beneficial when it is `Http`, as it allows hyper to make use vectored IO for better throughput.


lucab:
@seanmonstar it looks like trait signature doesn't match.

seanmonstar:
Yea, I goofed, and was also lazy and let CI build and test.

PR #78, opened 2019-08-19
bound the inner stream on AsyncRead + AsyncWrite rather than Read + Write

jbg:


ctz:
Thanks I will merge this and release 0.17.1 tonight

lucab:
I've also verified that this fixes the CI failure at https://github.com/seanmonstar/reqwest/pull/593.

Sorry again for my screw-up.

PR #79, opened 2019-08-19
Upgrade to std futures and tokio 0.2

dbcfd:
Upgrade to std futures and tokio 0.2


dbcfd:
https://github.com/ctz/hyper-rustls/compare/master...dbcfd:futures?expand=1#diff-2763e33724cc9b239c750ae973480953R76 

May need to do Pin::map_unchecked_mut here

PR #83, opened 2019-10-04
Update Dependencies to support Rust Stable

CryZe:
Since async await is now stable on Rust Beta 1.39, using the futures ecosystem should work just fine, except there's still many crates that have #![feature(async_await)] somewhere in their dependency graph. Most crates have been updated by now. This updates hyper-rustls as well.

### DIFF
--- a/examples/server.rs
+++ b/examples/server.rs
@@ -2,8 +2,12 @@
 //!
 //! First parameter is the mandatory port to use.
 //! Certificate and private key are hardcoded to sample files.
-#![feature(async_await)]
-use futures::{Stream, StreamExt, TryFutureExt, TryStreamExt};
+use core::task::{Context, Poll};
+use futures_util::{
+    stream::{Stream, StreamExt},
+    try_future::TryFutureExt,
+    try_stream::TryStreamExt,
+};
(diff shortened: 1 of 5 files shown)


examples/server.rs
```diff
@@ -66,14 +70,33 @@ async fn run_server() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+    type Error = io::Error;
+
+    fn poll_accept(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context,
+    ) -> Poll<Option<Result<Self::Conn, Self::Error>>> {
+        Pin::new(&mut self.acceptor).poll_next(cx)
+    }
+}
+
```
CryZe: I'm not too sure if this is the best solution. Basically hyper doesn't really use Stream anymore, so they have alternative traits such as Accept that you need to implement. Maybe something like this should be in hyper-rustls directly and not in an example?

CryZe:
What's the status of this? This should be ready to be merged.

CryZe:
I guess we can consider this "Rust stable" support now. Ping?

ctz:
Thanks! Should be ready for the cascade of async releases when 1.39 appears.

PR #85, opened 2019-10-13
Upgrade 0.18 to the latest futures-preview

alex:


alex:
CI failures, but I don't think they're caused by this?

CryZe:
This is already done in #83

ctz:
Thanks for the work here, I took #83 which was a superset of this change.

PR #91, opened 2019-11-24
Server example: Configure ALPN to accept h2, http/1.1

fluffysquirrels:


fluffysquirrels:
That test seems flaky on Azure Pipelines only. Is this still OK to go ahead?

ctz:
Thanks!

Yes that test is flickery for some reason on azure. Haven't managed to get it to break in a similar way locally :(

PR #95, opened 2019-12-08
Use ALPN by default to negotiate for HTTP2

### LINKED ISSUE #94: Why is HTTP2 not the default?
It seems like neither the client nor the server use HTTP2 by default. The server example recently got updated to manually allow HTTP2, but the client example still only uses HTTP1.1. Which begs the question why you even need to do this manually in the first place. I feel like either there should be some helper in the HttpsConnector .with_http2() or it should just be the default. I may be misunderstanding something though.

So hyper-rustls doesn't do ALPN by default and hyper-tls doesn't do ALPN at all. I'm so confused.

CryZe:
Hyper supports HTTP2, so hyper-rustls should use ALPN to negotiate for HTTP2 by default.

Resolves #94

### DIFF
--- a/src/connector.rs
+++ b/src/connector.rs
@@ -28,6 +28,7 @@ impl HttpsConnector<HttpConnector> {
         let mut http = HttpConnector::new();
         http.enforce_http(false);
         let mut config = ClientConfig::new();
+        config.alpn_protocols = vec![b"h2".to_vec(), b"http/1.1".to_vec()];
         config.root_store = rustls_native_certs::load_native_certs()
             .expect("cannot access native cert store");
         config.ct_logs = Some(&ct_logs::LOGS);


CryZe:
The commit builds, Azure Pipelines just timed out for no reason.

alex:
Ten seconds of grepping https://github.com/hyperium/hyper-tls and I don't see ALPN being set there either. Do we think that's a bug? Or am I bad at grep? I'd assume hyper-rustls would basically match hyper-tls on behavior.

CryZe:
hyper-tls uses native-tls and the PR for ALPN there has been rejected as not all native APIs it uses have working ALPN (it's basically blocked by Windows and macOS afaik).

alex:
Oh noes :-( Do I understand correctly that, at the moment, hyper's HTTP/2 support is basically never used then?

CryZe:
Yeah for hyper-tls you have to use the force_http2 (or whatever it's called) option that basically just assumes the server can handle HTTP2. For hyper-rustls you have to manually use the From impl for the HttpsConnector to pass the ALPN settings (at least until this PR). However hyper-openssl does proper ALPN by default on all platforms (at the "price" of having to use OpenSSL on Windows and macOS as well).

alex:
Thanks, the ecosystem summary was incredibly valuable, I appreciate it

PR #98, opened 2020-01-07
Update example: Ignore errors instead of abort

simnalamburt:
You don't want to restart the sample server everytime it aborts :)


simnalamburt:
Closed by https://github.com/rustls/hyper-rustls/commit/9b7b1220f74de9b249ce2b8f8b922fd00074c53b

PR #100, opened 2020-01-07
don't depend on hyper/tcp feature by default

erickt:
On fuchsia, we don't support the mio and tokio runtime yet, and so in order to use hyper-rustls, we need to compile it with `default-features = false`. Unfortunately, in https://github.com/ctz/hyper-rustls/pull/96, @CryZe added a hard requirement for the hyper "tcp" feature, which pulls in the tokio runtime.

To restore our ability to use hyper-rustls, this patch migrates the "hyper/tcp" hard feature requirement to an optional "tcp" feature. Note that we don't need to change the default case, because the "hyper/runtime" feature already depends on "hyper/tcp".

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ repository = "https://github.com/ctz/hyper-rustls"
 bytes = "0.5.2"
 ct-logs = { version = "^0.6.0", optional = true }
 futures-util = "0.3.1"
-hyper = { version = "0.13.0", default-features = false, features = ["tcp"] }
+hyper = { version = "0.13.0", default-features = false }
 rustls = "0.16"
 tokio = "0.2.4"
 tokio-rustls = "0.12.1"


erickt:
I just checked, and it actually looks like we don't even need this feature for `cargo test --no-default-features`, so I removed the optional `tcp` feature.

CryZe:
Whoops, sorry about that ^^'

erickt:
@CryZe No worries! This isn't blocking us yet, I'm just doing prep work to update to hyper 0.13. Thanks for doing the upgrade!

erickt:
@ctz gentle ping :)

erickt:
Thanks @lucab! Would it be possible to also get a 0.19.1 release cut for this change?

lucab:
You'd need @ctz for publishing a new crate version. He usually batches GH activity around the weekend.

erickt:
Awesome, thanks!

PR #102, opened 2020-01-24
Update to allow use of webpki-roots where rustls-native-certs fails.

ZaneHannanAU:
This adds the option to use webpki-roots in the cases where rustls-native-certs fails or is impossible, for example `aarch64-linux-android` (<https://github.com/ctz/rustls-native-certs/issues/3> is not possible, as the keystore is unavailable as per <https://developer.android.com/training/articles/keystore>), `aarch64-unknown-none` (it literally doesn't exist), microcontroller classes, `x86_64-unknown-cloudabi`/`aarch64-unknown-cloudabi` (also doesn't exist), `wasm32-wasi` (needs embedded or local thing), etc.

~~This additionally adds a compile time check for handling root store without required feature.~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,8 @@ before_script:
 
 script:
   - cargo test
-  - cargo test --all-features
+  - cargo test --no-default-features --features native-tokio
+  - cargo test --no-default-features --features webpki-tokio
   - cargo test --no-default-features
   - bash -c 'if [[ "$TRAVIS_RUST_VERSION" == "$CLIPPY_RUST_VERSION" ]]; then
       cargo clippy -- -D warnings;
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,9 @@ tokio-runtime =  ["hyper/runtime", "ct-logs"]
 [[example]]
 name = "client"
 path = "examples/client.rs"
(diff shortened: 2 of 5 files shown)


ZaneHannanAU:
Okay, for the time being leaving out the `all-features` variant with regards to this:

```rust
#[cfg(all(
    feature = "tokio-runtime",
    any(not(feature = "rustls-native-certs"), feature = "webpki-roots"),
    any(not(feature = "webpki-roots"), feature = "rustls-native-certs")
))]
compile_error!(
    "Must enable exactly one of rustls-native-certs (default) or webpki-roots with tokio-runtime! (note: use `default-features = false' in a binary crate for one or other)"
);
```

ZaneHannanAU:
As an alternative, I have https://github.com/ZaneHannanAU/rustls-native-certs/tree/allow-webpki-roots-on-unsupported-platform set up.

```toml
[path.crates-io]
rustls-native-certs = { git = "https://github.com/ZaneHannanAU/rustls-native-certs", branch = "allow-webpki-roots-on-unsupported-platform" }
```

DoumanAsh:
To be honest considering that `rustls-native-certs` is not really working well it would be better if webpki-roots would be default option

ctz:
Thanks!

PR #107, opened 2020-04-01
Allow construction with HttpConnector and default ClientConfig (closes #67)

### LINKED ISSUE #67: Allow construction with HttpConnector and default ClientConfig
See for example the example code at [hyper-trust-dns](https://github.com/jbg/hyper-trust-dns).

If I want to supply an `HttpConnector`, for example to change the resolver, it seems that I also have to also supply a `ClientConfig`.

It could be useful to offer a `from(http_connector: T)` which uses the default rustls `ClientConfig`.

inikulin:


inikulin:
Something is wrong with the CI for MacOS - builds get cancelled right away. It's definitely unrelated to the PR.

kinosang:
maybe a `new_with_connector()` is more suitable.

inikulin:
@kinosang thanks for the suggestion and we looked into it, but we'd like to stick with `hyper-rustls` in our project.

kinosang:
@inikulin yeah, I mean, to uniformize the APIs, we may have a `new_with_connector()` for hyper-rustls instead of a `default_client_config`.

inikulin:
This is what I did initially, however later we've had a case there we needed to do minor adjustments to the default config and this API is flexible enough to support that.

aheadmode:
Wouldn't it be more sufficient to drop ~~`default_client_config`~~ which seems kinda awkward and introduce something like `HttpsClientConfig`? We could still use `new_with_connector` and keep the API aligned with [hyper-tls](https://github.com/hyperium/hyper-tls).

Example

```rust
use hyper_timeout_connector::HttpTimeoutConnector;
use hyper_rustls::{HttpsConnector, HttpsClientConfig};
use std::time::Duration;

let mut http_connector = HttpTimeoutConnector::new();
http_connector.set_connect_timeout(Some(Duration::from_secs(10)));

// Create a new HttpsClientConfig
let mut client_config = HttpsClientConfig::default();
client_config.foo = ...

let https_connector = HttpsConnector::from(http_connector, client_config);
```

```rust
use hyper_timeout_connector::HttpTimeoutConnector;
use hyper_rustls::HttpsConnector;
use std::time::Duration;

let mut http_connector = HttpTimeoutConnector::new();
http_connector.set_connect_timeout(Some(Duration::from_secs(10)));

// Uses HttpsClientConfig by default
let https_connector = HttpsConnector::new_with_connector(http_connector);
```

petabyteboy:
@inikulin please rebase

@lucab  @ctz could you have a look at this PR? I need exactly the same thing and this looks like an easy change.

PR #108, opened 2020-04-26
added HttpsConnecor::new_with_certificate_verifier

vproxy666:
Added this method to allow user to customize the certificate verifier


vproxy666:
Maybe it is better to add `fn new(cfg:ClientConfig)` so that user can totally control it?

PR #111, opened 2020-06-28
Remove brackets of ipv6 host address.

fbzhong:
Remove brackets of ipv6 host address, which is introduced by https://github.com/hyperium/http/pull/292


ctz:
I believe this crate won't work with literal IP addresses as yet; so I'm a bit confused by this PR.

fbzhong:
@ctz Yes, it's true that https://github.com/briansmith/webpki does not support iPAddress SAN, but I believe it's just matter of time, it will eventually support it.

djc:
I think this was fixed in #182.

PR #114, opened 2020-08-07
Made server example tls_acceptor comment more clear

jspspike:
Previous comment would not prevent server abort. Changed to use `filter_map` and updated comment.


cpu (approved):
This looks like a reasonable improvement to me and perhaps a better alternative to https://github.com/rustls/hyper-rustls/pull/98. @djc WDYT?

PR #133, opened 2020-11-27
Enable hyper client feature

messense:
The HTTP client of hyper is now an optional feature: https://github.com/hyperium/hyper/commit/4e55583d30a597884883f1a51b678f5c57c76765


messense:
Not sure why the CI failed.

ctz:
Thanks! (I've adjusted this to the "full" feature, since "client" also requires "http1" or "http2" to do anything.)

PR #134, opened 2020-11-28
Force backend choice through API rather than features

djc:
Unlike the suggestion in #125, this only enables the native-tokio feature by default, but it enables all features in docs.rs, annotating both of the `HttpsConnector` constructors with the required feature flags (the required feature is only available in nightly for now, but docs.rs always runs nightly rustdoc). I believe this is better since it doesn't unnecessarily bloat the dependency graph of downstream crates, but I'm happy to change it if you feel strongly about enabling both.


ctz:
Thanks!

PR #139, opened 2020-12-29
Don't enable the hyper http1 feature now that hyper 0.14.2 is out

paolobarbolini:
Starting from hyper 0.14.2 http connection types don't require a specific proto feature to be enabled in order to be accessed.

See https://github.com/hyperium/hyper/pull/2377

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ repository = "https://github.com/ctz/hyper-rustls"
 log = "0.4.4"
 ct-logs = { version = "^0.8", optional = true }
 futures-util = "0.3.1"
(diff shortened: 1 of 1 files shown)


CryZe:
Uhm, okay so this is all very weird. We always send the ALPN protocols h2 and http/1.1 which means the server is free to choose either, meaning that technically hyper-rustls requires hyper to support both these protocols.

paolobarbolini:
If ALPN is always sent then we are already doing it wrong, since we only enable `http1`

CryZe:
Yeah that's what I noticed as well.

djc:
I think #144 effectively mitigates this, right?

paolobarbolini:
Right. Closing

PR #142, opened 2021-01-08
Remove futures-util

paolobarbolini:
`h2`, `hyper` and `reqwest` import `futures-util` with no default features, to make it faster to compile. This crate doesn't actually need it, so I moved it to dev-dependencies where an example uses it.


jplatte:
Ping @ctz

lucab (approved):

PR #144, opened 2021-02-10
fix: Only report http2 support if the http2 feature is enabled

Marwes:
Fixes https://github.com/ctz/hyper-rustls/issues/143

### DIFF
--- a/src/connector.rs
+++ b/src/connector.rs
@@ -65,10 +65,6 @@ impl HttpsConnector<HttpConnector> {
         {
             config.alpn_protocols = vec![b"h2".to_vec(), b"http/1.1".to_vec()];
         }
-        #[cfg(not(feature = "http2"))]
-        {
-            config.alpn_protocols = vec![b"http/1.1".to_vec()];
-        }
 
         config.ct_logs = Some(&ct_logs::LOGS);
         (http, config).into()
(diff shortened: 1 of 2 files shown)


src/connector.rs
```diff
@@ -61,7 +61,15 @@ impl HttpsConnector<HttpConnector> {
...
-        config.alpn_protocols = vec![b"h2".to_vec(), b"http/1.1".to_vec()];
+        #[cfg(feature = "http2")]
+        {
+            config.alpn_protocols = vec![b"h2".to_vec(), b"http/1.1".to_vec()];
+        }
+        #[cfg(not(feature = "http2"))]
+        {
+            config.alpn_protocols = vec![b"http/1.1".to_vec()];
+        }
+
```
g2p, line 72: I think alpn_protocols should just be omitted if it only contains "http/1.1".

Arnavion:
`hyper` allows toggling `http1` and `http2` features independently. So you also need to provide a knob for the ALPN negotiation to only include `"h2"` and not `"http/1.1"` for the case where the client only supports HTTP/2 and not /1.1

paolobarbolini:
Any updates on this?

djc:
This makes sense to me, thanks!

Issue #143, opened 2021-02-03
Don't force HTTP2 support


g2p:
Currently hyper-rustls sets alpn_protocols to advertise HTTP2 (in connector.rs).
It doesn't properly work with current hyper features (which require http2 to be explicitly enabled), and forces the user to enable a feature that has a large dependency footprint and potential security surface (or to be incompatible with some peers).

hyper-rustls should give the caller some control over that, prior to building the connector.
And the support should be conditional on an hyper-rustls/http2 feature, so that the dependency set is always correct at compile time. It could work similar to how the root certificates are determined.

Arnavion:
To save other people the time to investigate this: If you're using a `hyper::Client`, and you only enabled the `http1` feature of `hyper` and not the `http2` feature, and you see a `channel closed` error from hyper from executing some (but not necessarily all) of your client requests, this issue may be why. To be sure, enable trace logging for rustls and see if there's a log indicating it selected HTTP/2 during ALPN right before the request failed:

>[DEBUG rustls::client::hs] ALPN protocol is Some(b"h2")

For requests that succeed, you'd instead see `ALPN protocol is None` or `ALPN protocol is Some(b"http/1.1")`

kahlil29:
Just to clarify, if I'm using the `rusoto` crate with the `rustls` feature enabled (I think it depends on `hyper-rustls`), is there a way for me to know if I'm hitting this issue? 
i.e. (Can I) How do I enable logging that would enable me to narrow down the issue or confirm that this issue affects me?

PR #145, opened 2021-02-11
Add HttpsConnectorBuilder

g2p:
This gives more control over various rustls features,
as well as ensures that enabling connector features
like http2 can only be done when the appropriate crate
features are enabled.

Adds an hyper-rustls/http2 feature, fixing #143.

This is an alternative to the fix in #144, my motivation is that crate features can be activated at a remote whereas an explicit method call removes ambiguity.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,9 +28,10 @@ futures-util = { version = "0.3.1", default-features = false }
 
 [features]
 default = ["native-tokio"]
+http2 = ["hyper/http2"]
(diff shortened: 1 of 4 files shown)


paolobarbolini:
I didn't see this when it was originally submitted but I agree this is better than doing it through a feature.

djc:
Rebasing this on top of #153 makes sense, I think (I can now also merge PRs in this project).

djc:
With #156, I think this is no longer needed, right?

g2p:
Right!

PR #147, opened 2021-03-22
Use server example based on warp tls

pickfire:
The current server example is not able to process another request
while tls is resolving the cert, which prevents acme client from
solving acme tls challenges, similarly during that all other
requests will be blocked. Although I am not quite sure why but
the tls in warp worked out for me.

I forgot what I changed but I recall I was trying testing out with the new resolver but I wonder why it can't seemed to receive challenge request from acme, then I noticed that it can't even receive requests from another curl client, then I found the issue out.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,6 @@ webpki = "0.21.0"
 webpki-roots = { version = "0.21", optional = true }
 
 [dev-dependencies]
-async-stream = "0.3.0"
 tokio = { version = "1.0", features = ["io-std", "macros", "net", "rt-multi-thread"] }
 hyper = { version = "0.14", features = ["full"] }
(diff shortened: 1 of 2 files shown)


djc:
Sorry for the delay in reviewing, but this looks reasonable to me. Would you be interested in rebasing this on top of main?

pickfire:
Done

djc:
Would you mind adding a separate commit bumping the MSRV to 1.51?

pickfire:
> Would you mind adding a separate commit bumping the MSRV to 1.51?

How should I do that?

djc:
Yeah, just replace the `1.50` in the workflow file with `1.51`.

djc:
Thanks!

PR #153, opened 2021-07-06
Upgrade to rustls 0.20

djc:
This is blocked on tokio-rustls and rustls-native certs releasing updates, first.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,14 +11,15 @@ repository = "https://github.com/ctz/hyper-rustls"
 
 [dependencies]
 log = "0.4.4"
-ct-logs = { version = "^0.8", optional = true }
+ct-logs = { version = "^0.9", optional = true }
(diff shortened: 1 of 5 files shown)


g2p:
There's a remaining issue with the released version of rustls, this doesn't build as is because rustls now forces users to figure out how long they trust a list of logs to be current (after which the default verifier fails open), while rustls is a bit more lax than Chrome (checking just one log), users expecting hyper-rustls to connect to any server that Chrome will connect to would need hyper-rustls to embed at build time a time stamp from when the list of ct-logs was last checked to be current.
I don't think we can do that, plus the ct-logs crate isn't current at the moment, so I don't think we can enable sct validation out of the box like before.

djc:
I agree that for now it's probably better to keep CT validation disabled for now.

djc:
Finished off in #156.

PR #155, opened 2021-09-24
Add `force_https` flag which enforces https only connections.

boxdot:


paolobarbolini:
This would be great on top of #145

boxdot:
> This would be great on top of #145

Indeed. I'll wait until it is merged and rebase then.

djc:
LGTM, thanks!

PR #160, opened 2021-11-16
Fix bad assert on alpn_protocols and add tests

g2p:
Remove the second assert added in 54e0757e938ffc5b490d49a12c8a6155dbf72745, keep only the first.
Add tests.

An assert was added on WantsProtocols1.wrap_connector which was the wrong place, as we update the alpn_protocols field ourselves before calling wrap_connector.
Asserting on with_tls_config is sufficient, this is the only place ConnectorBuilder(WantsSchemes) is constructed and the builder flow will always go through here.

Add tests for all expected values of alpn_protocols and for the expected panic when trying to force it from outside.


djc:
Thanks for fixing that up!

PR #163, opened 2022-01-03
Document dependency on `http2` feature

tgeoghegan:
Annotates the rustdoc on `ConnectorBuilder<WantsProtocols{1,2}>::enable_http2`
so that docs.rs will plainly tell users that crate feature `http2` must
be enabled.


tgeoghegan:
Preview of what this looks like on Firefox:
image

djc (commented):
Thanks, that's helpful!

djc (approved):

Issue #170, opened 2022-05-03
how should I configure features to support http1, http2 and tls?


garypen:
Thanks for working on this crate. It's really useful. I do have an issue though:

If I add hyper-rustls to my Cargo.toml as follows:

```
hyper-rustls = { version = "0.23.0" }
```

I find that I can't compile my code:

```
        let connector = hyper_rustls::HttpsConnectorBuilder::new()
            .with_native_roots()
            .https_or_http()
            .enable_http2()
            .enable_http1()
            .build();
```

because I get this error:

```
error[E0599]: no method named `enable_http2` found for struct `HttpsConnectorBuilder` in the current scope
  --> apollo-router-core/src/services/tower_subgraph_service.rs:31:14
   |
31 |             .enable_http2()
   |              ^^^^^^^^^^^^ help: there is an associated function with a similar name: `enable_http1`
```

Reading through the documentation, I find references to features and modify my Cargo.toml so that:

```
hyper-rustls = { version = "0.23.0", features = ["http1", "http2", "tls12"] }
```

I then get this error:

```
error[E0599]: no method named `enable_http1` found for struct `HttpsConnectorBuilder<WantsProtocols3>` in the current scope
  --> apollo-router-core/src/services/tower_subgraph_service.rs:32:14
   |
32 |             .enable_http1()
   |              ^^^^^^^^^^^^ method not found in `HttpsConnectorBuilder<WantsProtocols3>`
   |
   = note: the method was found for
           - `HttpsConnectorBuilder<WantsProtocols1>`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `apollo-router-core` due to previous error
```

Should it be possible to have both `http2` and `http1` enabled?

paolobarbolini:
Have you tried swapping `.enable_http2()` and `.enable_http1()`?

garypen:
I thought I had, but apparently I hadn't. I just swapped them around and this compiles fine:
```
hyper-rustls = { version = "0.23.0", features = ["http1", "http2", "tls12"] }
```

```
        let connector = hyper_rustls::HttpsConnectorBuilder::new()
            .with_native_roots()
            .https_or_http()
            .enable_http1()
            .enable_http2()
            .build();
```

I guess I should have figured that out from the documentation somehow, but is there a way this requirement could be made clearer? Perhaps in the docs for both `enable_http1` and `enable_http2`?

djc:
Proposed an additional helper method in #171 that might make this nicer.

[...]

PR #171, opened 2022-05-09
Add ConnectorBuilder::enable_all_versions() helper

djc:
See #170. I remember also getting confused about this recently.


cpu:
This looks like a nice/small change. If it was rebased on main perhaps we could get it merged :crossed_fingers:

cpu (approved):
This looks like a nice change with small scope. I'm comfortable +1ing it.

PR #177, opened 2022-09-17
Allow overriding server name

MikailBag:
Original case: kube-rs/kube-rs#991

If I understand correctly, hyper-rustls (and other TLS connectors) is the only place to implement this feature since it is responsible for calling [this](https://docs.rs/rustls/latest/rustls/client/struct.ClientConnection.html#method.new) function.


MikailBag:
CI failure seems unrelated

src/connector.rs
```diff
@@ -83,21 +85,28 @@ where
...
+                    }
+                };
                 let connecting_future = self.http.call(dst);
 
                 let f = async move {
                     let tcp = connecting_future
                         .await
                         .map_err(Into::into)?;
                     let connector = TlsConnector::from(cfg);
-                    let dnsname = rustls::ServerName::try_from(hostname.as_str())
```
MikailBag, line 97: This is unrelated change, but parsing hostname outside of the async block saves us one allocation.

src/connector.rs
```diff
@@ -21,6 +21,7 @@ pub struct HttpsConnector<T> {
     force_https: bool,
     http: T,
     tls_config: Arc<rustls::ClientConfig>,
+    server_name_override: Option<String>,
```
djc, line 24: I'd prefer to call this `override_server_name`.

src/connector.rs
```diff
@@ -83,21 +85,28 @@ where
...
+                    Err(_) => {
+                        let err = io::Error::new(io::ErrorKind::Other, "invalid dnsname");
+                        let err: Box<dyn std::error::Error + Send + Sync + 'static> = Box::new(err);
+                        let res = std::future::ready(Err(err));
+                        return Box::pin(res);
+                    }
+                };
                 let connecting_future = self.http.call(dst);
 
                 let f = async move {
```
djc, line 103:
```suggestion
                };

                let hostname = match rustls::ServerName::try_from(hostname) {
                    Ok(dnsname) => dnsname,
                    Err(_) => {
                        let err = io::Error::new(io::ErrorKind::Other, "invalid DNS name");
                        return Box::pin(std::future::ready(Err(Box::new(err) as BoxError)));
                    }
                };

                let connecting_future = self.http.call(dst);
                let f = async move {
```

src/connector/builder.rs
```diff
@@ -169,6 +173,15 @@ impl ConnectorBuilder<WantsProtocols1> {
             enable_http1: false,
         })
     }
+
+    /// Override expected server name
```
djc, line 177: I think you should pick one stage where this can be set, instead of adding it to all three. Please rename the method to `with_server_name()` and explain in the documentation that this overrides the server name derived from the destination URL's host name.

djc (commented):
Thanks, this broadly makes sense. Some improvements suggested below.

src/connector/builder.rs
MikailBag: I left only one insance, at the `WantsProtocols1` stage

src/connector.rs
```diff
@@ -83,21 +85,26 @@ where
...
-                    .to_string();
+                let hostname = match self.override_server_name.as_deref() {
+                    Some(h) => h,
+                    None => dst.host().unwrap_or_default(),
+                };
+                let hostname = match rustls::ServerName::try_from(hostname) {
+                    Ok(dnsname) => dnsname,
+                    Err(_) => {
+                        let err = io::Error::new(io::ErrorKind::Other, "invalid dnsname");
+                        return Box::pin(std::future::ready(Err(Box::new(err) as BoxError)));
```
djc, line 96: Would prefer to use `async move { Err(err.into()) }` instead of `std::future::ready()` here.

src/connector/builder.rs
```diff
@@ -194,7 +212,7 @@ impl ConnectorBuilder<WantsProtocols2> {
             enable_http1: true,
         })
     }
-
+
```
djc, line 215: Please ditch the trailing whitespace.

djc (commented):
Thanks! Some minor nits left...

MikailBag:
Thanks!
It'd be nice to see a new release later this month, if comfortable.

PR #182, opened 2022-12-05
fix: can't resolve ip address with square brackets

digitwolf:
Problem:
Sending to destination like https://[::1]/ fails with error invalid dnsname. Error is thrown here https://github.com/rustls/hyper-rustls/blob/main/src/connector.rs#L95

hyper-rustls resolves the Hostname as hyper::Uri::host This makes hostname = "[::1]" which is HTTP-specific form of writing IPv6 address.

Solution:
Since square brakets are coming from hyper and not part standard IP notation we remove any square brakets inside of the connector before calling the rustls::ServerName::try_from(hostname)

### DIFF
--- a/src/connector.rs
+++ b/src/connector.rs
@@ -85,10 +85,19 @@ where
                 Box::pin(f)
             } else if sch == &http::uri::Scheme::HTTPS {
                 let cfg = self.tls_config.clone();
-                let hostname = match self.override_server_name.as_deref() {
+                let mut hostname = match self.override_server_name.as_deref() {
                     Some(h) => h,
                     None => dst.host().unwrap_or_default(),
                 };
+
+                // Remove square brackets around IPv6 address.
+                if let Some(trimmed) = hostname
+                    .strip_prefix('[')
+                    .and_then(|h| h.strip_suffix(']'))
+                {
+                    hostname = trimmed;
+                }
+
                 let hostname = match rustls::ServerName::try_from(hostname) {
                     Ok(dnsname) => dnsname,
                     Err(_) => {


src/connector.rs
```diff
@@ -85,10 +85,16 @@ where
...
             } else if sch == &http::uri::Scheme::HTTPS {
                 let cfg = self.tls_config.clone();
-                let hostname = match self.override_server_name.as_deref() {
+                let mut hostname = match self.override_server_name.as_deref() {
                     Some(h) => h,
                     None => dst.host().unwrap_or_default(),
                 };
+
+                // Remove square brackets around IPv6 address.
+                if hostname.starts_with('[') && hostname.ends_with(']') && hostname.len() > 2 {
```
djc, line 94: Any reason we can't use [`trim_matches()`](https://doc.rust-lang.org/std/primitive.str.html#method.trim_matches) here?

Cargo.toml
```diff
@@ -14,15 +14,15 @@ http = "0.2"
 hyper = { version = "0.14", default-features = false, features = ["client"] }
 log = { version = "0.4.4", optional = true }
 rustls-native-certs = { version = "0.6", optional = true }
-rustls = { version = "0.20.1", default-features = false }
+rustls = { version = "0.20.7", default-features = false }
```
djc, line 17: This commit doesn't help, versions in `Cargo.toml` allow the use of semantically compatible new versions by default (and in fact, Cargo will prefer newer versions), so this change is meaningless here.

src/connector.rs
digitwolf: I wasn't aware of this Rust API. Can update this to 
```
hostname = hostname.trim_matches(|c| c == '[' || c == ']')
```

CryZe: That seems worse as you really only want to trim [ on the left and ] on the right.

digitwolf: Good point!
What about
```
hostname = hostname.trim_start_matches('[').trim_end_matches(']')
```
?

src/connector.rs
```diff
@@ -85,10 +85,16 @@ where
...
+                    hostname = &hostname[1..hostname.len() - 1];
```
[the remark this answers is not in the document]
digitwolf: Applied your suggestion!
Thank you!

Cargo.toml
djc: It's explained here: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio.

djc (approved):

digitwolf:
djc, I think I hit the wrong button, please approve the workflow again

djc:
Okay, thanks!

PR #184, opened 2022-12-16
tests: support CARGO_TARGET_DIR env variable

MarcoIeni:
Tests don't work if you have a custom `CARGO_TARGET_DIR`

### DIFF
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -1,10 +1,35 @@
+use std::env;
+use std::path::PathBuf;
 use std::process::Command;
 use std::thread;
 use std::time;
 
+fn target_dir() -> PathBuf {
+    env::var("CARGO_TARGET_DIR")
+        .unwrap_or_else(|_| "target".to_string())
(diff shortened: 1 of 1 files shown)


tests/tests.rs
```diff
@@ -1,10 +1,35 @@
...
+}
+
+fn server_command() -> Command {
+    let server_bin_path = examples_dir().join("server");
+    Command::new(server_bin_path)
+}
+
+fn client_command() -> Command {
+    let client_bin_path = examples_dir().join("client");
+    println!("client_bin_path: {:?}", client_bin_path);
```
djc, line 26: Can we remove the `println!()` call?

I'd prefer to inline `target_dir()` into `examples_dir()` and avoid the bindings in `server_command()` and `client_command()` (instead putting the `examples_dir().join()` directly in the `Command::new()` call.

djc (commented):
Thanks!

djc (approved):

PR #191, opened 2023-02-07
Add `HttpsConnector::https_only` for compatibility with `hyper_tls`

jjant:
## Description
It's convenient to be able to configure this flag after building the `HttpsConnector`, and [`hyper_tls` has this functionality](https://docs.rs/hyper-tls/latest/hyper_tls/struct.HttpsConnector.html#method.https_only) so this is nice for compatibility.

See https://github.com/awslabs/smithy-rs/pull/2316#discussion_r1097744060 for motivation.

### DIFF
--- a/src/connector.rs
+++ b/src/connector.rs
@@ -23,6 +23,15 @@ pub struct HttpsConnector<T> {
     override_server_name: Option<String>,
 }
 
+impl<T> HttpsConnector<T> {
+    /// Force the use of HTTPS when connecting.
+    ///
+    /// If a URL is not `https` when connecting, an error is returned.
+    pub fn https_only(&mut self, enable: bool) {
+        self.force_https = enable;
+    }
+}
+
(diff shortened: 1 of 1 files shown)


djc:
So I don't mind the use case of enabling `force_https` after construction, but this API can also be used to do the opposite, which seems potentially trickier? Like, any caller with a `&mut HttpsConnector` can now stop enforcing the use of HTTPS, which might be the opposite of the caller intentions when passing in a fully-formed `HttpsConnector`.

If you change this to making it unidirectional (maybe `enforce_https()`?), I don't see any issues.

jjant:
Thanks for the suggestion, @djc, I've made the change in https://github.com/rustls/hyper-rustls/pull/191/commits/6f83fc5f151f0e1f86ec5f5dd53c846b5190c886.

djc:
Thanks!

PR #197, opened 2023-03-15
CI: replace ubuntu 18.04 w/ 20.04.

cpu:
Quick follow-up to #196.

The Build+Test matrix job of the GitHub actions workflow was using Ubuntu 18.04, which is deprecated.

This commit replaces it with Ubuntu 20.04, the oldest supported Ubuntu release available in GitHub actions at the time of writing.

The other workflow task are using `ubuntu-latest` and do not need updating at this time.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ jobs:
           - beta
           - nightly
           - 1.57
-        os: [ubuntu-18.04]
+        os: [ubuntu-20.04]
         # but only stable on macos/windows (slower platforms)
         include:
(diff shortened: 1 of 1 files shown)


cpu:
> cpu force-pushed the cpu-rm-ubuntu-18.04 branch from 8ee5639 to f9aa48e 

Changed to update from 18.04 to 20.04 instead of 22.04 based on feedback from [djc on another PR](https://github.com/rustls/pemfile/pull/16#discussion_r1137483370).

djc (approved):

PR #204, opened 2023-04-10
Add documentation metadata to Cargo.toml so crates.io will link directly to it.

khuey:


cpu (approved):
I was initially confused because [the crates homepage for `hyper-rustls`](https://crates.io/crates/hyper-rustls) _does_ link to the documentation already but I was able to find some other places in the UI (notably search results) where having this metadata resulted in a direct link that wasn't present otherwise. Seems worthwhile :+1: Thanks!

khuey:
Yeah, it's the search results this adds the link to, sorry for not being more specific.

PR #205, opened 2023-05-24
Only check library against MSRV

djc:


djc:
I'll just merge this, so you can get the PR in the other one passing.

Issue #198, opened 2023-03-22
Cannot access peer certificates with example's TlsStream 


moshec2:
In the example's TlsStream, you cannot access the peer certificates from tokio_rustls's TlsStream in `make_service_fn`, even if you try to expose the state property, it's basically always Handshaking. Is there some way to change the example so you could wait until the handshaking is over so you can access this?

moshec2:
I will add some info on what I tried. One of the things I tried is exposing the state object in the hopes that inside `make_service_fn` it would be `Streaming` and I could get the peer certificates from the server connection, but sadly it's always Handshaking (which makes sense since it only does the handshake on a read/write):
https://gist.github.com/moshec2/8b1c99396f61e4b132412468fdb5e2c2#file-server-rs-L103
https://gist.github.com/moshec2/8b1c99396f61e4b132412468fdb5e2c2#file-server-rs-L68
Then I tried forcing the handshake to occur before and make the acceptor return the straight up `tokio_rustls::server::TlsStream` object, by creating the `tokio_rustls::Accept` inside the custom acceptor and polling it immediately:
```rs
impl Accept for TlsAcceptor {
    // type Conn = TlsStream;
    type Conn = tokio_rustls::server::TlsStream<AddrStream>;
    type Error = io::Error;

    fn poll_accept(
        self: Pin<&mut Self>,
        cx: &mut Context<'_>,
    ) -> Poll<Option<Result<Self::Conn, Self::Error>>> {
        let pin = self.get_mut();
        match ready!(Pin::new(&mut pin.incoming).poll_accept(cx)) {
            Some(Ok(sock)) => {
                let mut accept = tokio_rustls::TlsAcceptor::from(pin.config.clone()).accept(sock);

                match ready!(Pin::new(&mut accept).poll(cx)) {
                    Ok(stream) => Poll::Ready(Some(Ok(stream))),
                    Err(e) => Poll::Ready(Some(Err(e))),
                }

                // Poll::Ready(Some(Ok(TlsStream::new(sock, pin.config.clone()))))
            }
            Some(Err(e)) => Poll::Ready(Some(Err(e))),
            None => Poll::Ready(None),
        }
    }
}
```
Which reads like it would work, and after changing my `make_service_fn` accordingly this compiles but completely breaks the HTTPS server, the first connection ends abruptly during the handshake and then the server hangs indefinitely.

dydokamil:
[...]

PR #207, opened 2023-06-13
Enable access to acceptor::TlsStream internals

djc:


cpu:
Seems reasonable if it addresses the needs of the folks that chimed in on #198 :+1:

PR #213, opened 2023-07-24
Disable unnecessary default features from futures-util

djc:
futures-util recently grew some more dependencies (via futures-macros) which we don't need here.


cpu (approved):
Thanks!

PR #219, opened 2023-08-14
Acceptor improvements

### LINKED ISSUE #218: Getting ip address of connection
After switching to `hyper_rustls` from `hyper` input for `make_service_fn` is changed from `AddrStream` to `acceptor::TlsStream` which is not exposed as public API and doesn't have `remote_addr` function from the original struct, is there any way to get address of current connection/stream?

djc:
Fixes #218 and maybe also addresses #198?

### DIFF
--- a/src/acceptor.rs
+++ b/src/acceptor.rs
@@ -16,6 +16,52 @@ mod builder;
 pub use builder::AcceptorBuilder;
 use builder::WantsTlsConfig;
 
+/// A TLS acceptor that can be used with hyper servers.
+pub struct TlsAcceptor {
+    config: Arc<ServerConfig>,
+    incoming: AddrIncoming,
+}
+
+/// An Acceptor for the `https` scheme.
+impl TlsAcceptor {
+    /// Provides a builder for a `TlsAcceptor`.
+    pub fn builder() -> AcceptorBuilder<WantsTlsConfig> {
+        AcceptorBuilder::new()
+    }
+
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -83,7 +83,7 @@ jobs:
       - name: Install rust toolchain
         uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: "1.60"
+          toolchain: "1.63"
```
cpu, line 86: It's a little bit unfortunate to have this crate use a more aggressive MSRV than the Rustls + Webpki crates :-/

The part that has me a little bit confused is that it seems like we're pulling in socket2 from hyper, and hyper's MSRV for the 0.14.x release is lower than 1.60.

Am I missing something?

cpu (approved):
Thanks for the many cleanup commits (and adding cron CI!).

This looks good to me. I had one question about the MSRV increase but I don't think it's blocking.

.github/workflows/build.yml
djc: socket2 0.5.3 is being pulled in by Tokio 1.30+. rustls and webpki don't depend on tokio, IIRC? I bumped the tokio-rustls MSRV last week, too: https://github.com/rustls/tokio-rustls/pull/10.

webpki and rustls are just being conservative. That tokio is making the jump might mean that there's not much benefit anymore to keeping the MSRV below 1.63, but IMO the benefit to going to 1.63 is relatively limited for those crates.

cpu: Makes sense, thanks

PR #221, opened 2023-09-15
add TlsAcceptor::with_acceptor method

MarinPostma:
This PR add a `AcceptorBuilder::with_connector` method that takes a generic acceptor implementing `Accept`.
This is necessary when you need to work with any other acceptor than `AddrIncomming`.

### DIFF
--- a/src/acceptor.rs
+++ b/src/acceptor.rs
@@ -17,9 +17,9 @@ pub use builder::AcceptorBuilder;
 use builder::WantsTlsConfig;
 
 /// A TLS acceptor that can be used with hyper servers.
-pub struct TlsAcceptor {
+pub struct TlsAcceptor<A = AddrIncoming> {
     config: Arc<ServerConfig>,
-    incoming: AddrIncoming,
+    acceptor: A,
 }
 
 /// An Acceptor for the `https` scheme.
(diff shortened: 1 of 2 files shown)


djc (commented):
This mostly makes sense to me. I'd prefer to keep the bounds in the `impl` definition where possible (`impl<C: AsyncRead + AsyncWrite + Unpin>` TlsStream<c>`).

Do you have a sense of whether this is semver-compatible? If not, can you add a semver-breaking version bump in a separate commit?

Out of curiosity, what do you (want to) use instead of `AddrIncoming`?

MarinPostma:
@djc thanks, I will do as suggested

We're abstracting networking in our project to implement simulation with turmoil. So I have a custom acceptor that wraps a turmoil `TcpStream`

MarinPostma:
as for semver, I'm not completely certain. It seems to me that yes, thanks to the default types, but I'm not an expert enough on semver.

djc:
I think [this documentation](https://doc.rust-lang.org/cargo/reference/semver.html#generic-new-default) says it should be okay.

MarinPostma:
@djc, I fixed fmt and fixed the trait bounds 👍

djc (approved):
LGTM, thanks!

djc:
@cpu do you want to take a look?

cpu (approved):
Thanks, LGTM

PR #224, opened 2023-10-03
Connector: `enable_all_versions` with matching alpn vs features

### LINKED ISSUE #208: `HttpsConnectorBuilder::enable_all_versions` doesn't enable ALPN for http/1.1
When building a `HttpsConnectorBuilder` using `enable_all_versions`, ALPN does not seem to be enabled with `http/1.1`, and only `h2` is enabled:

https://github.com/rustls/hyper-rustls/blob/286e1fa57ff5cac99994fab355f91c3454d6d83d/src/connector/builder.rs#L183

However, calling `enable_http1` followed by `enable_http2` does enable ALPN for both versions.

This causes the client example to fail when using a `http/1.1` server, and a client built with `enable_all_versions`.

For example, with https://github.com/mmalenic/hyper-rustls/commit/863812f617c3b752ee4b32dcd13456be41053692, running the following:
```sh
cargo run --example server
```
```sh
cargo run --example client --features http2 "https://localhost:1337" examples/sample.pem
```
Results in the error:
```
FAILED: Could not get: hyper::Error(Connect, Custom { kind: Other, error: Custom { kind: InvalidData, error: AlertReceived(NoApplicationProtocol) } })
```
Is this the intended behaviour?

cc @norling @dbampalikis @brainstorm

TimLuq:
Fixes #208

The original intent of this function is slightly confusing. As it was implemented before, it required *all* supported protocol features to be enabled for the function to exist (both `http1` and `http2`). My naive guess would have been that the expected behavior is that all *enabled* protocol features would inform the actual functionality, but the function would still show up if at least one protocol is enabled.

I understand that having this completely dynamic based on active features would muddle the documented return type and be breaking depending on active features. Either way, to me, it hurt a bit working on a function named `enable_all_versions` with a strict feature requirement on the non-default optional feature `http2`. But now it should work correctly in the previously intended use cases.

### Changes to `enable_all_versions`
- drop strict requirement of the `http1` feature
- add feature requirement `http2` to fn docs (previously a non-documented requirement)
- alpn will only contain `h2` if `http1` is not enabled
- alpn will contain `h2` and `http/1.1` if both features are enabled

### DIFF
--- a/src/connector/builder.rs
+++ b/src/connector/builder.rs
@@ -174,7 +174,7 @@ impl ConnectorBuilder<WantsProtocols1> {
         })
     }
 
-    /// Enable all HTTP versions built into this library
+    /// Enable all HTTP versions built into this library (enabled with Cargo features)
     ///
     /// For now, this could enable both HTTP 1 and 2, depending on active features.
     /// In the future, other supported versions will be enabled as well.
@@ -350,5 +350,27 @@ mod tests {
             &connector.tls_config.alpn_protocols,
             &[b"h2".to_vec(), b"http/1.1".to_vec()]
         );
+    }
+
+    #[test]
+    #[cfg(all(not(feature = "http1"), feature = "http2"))]
+    fn test_alpn_http2() {
+        let roots = rustls::RootCertStore::empty();
+        let tls_config = rustls::ClientConfig::builder()
+            .with_safe_defaults()
+            .with_root_certificates(roots)
+            .with_no_client_auth();
+        let connector = super::ConnectorBuilder::new()
+            .with_tls_config(tls_config.clone())
+            .https_only()
+            .enable_http2()
+            .build();
+        assert_eq!(&connector.tls_config.alpn_protocols, &[b"h2".to_vec()]);
+        let connector = super::ConnectorBuilder::new()
+            .with_tls_config(tls_config)
+            .https_only()
+            .enable_all_versions()
+            .build();
+        assert_eq!(&connector.tls_config.alpn_protocols, &[b"h2".to_vec()]);
     }
 }
--- a/src/connector/builder.rs
+++ b/src/connector/builder.rs
@@ -174,16 +174,22 @@ impl ConnectorBuilder<WantsProtocols1> {
         })
     }
 
-    /// Enable all HTTP versions
+    /// Enable all HTTP versions built into this library
     ///
-    /// For now, this enables both HTTP 1 and 2. In the future, other supported versions
-    /// will be enabled as well.
-    #[cfg(all(feature = "http1", feature = "http2"))]
+    /// For now, this could enable both HTTP 1 and 2, depending on active features.
+    /// In the future, other supported versions will be enabled as well.
+    #[cfg(feature = "http2")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "http2")))]
     pub fn enable_all_versions(mut self) -> ConnectorBuilder<WantsProtocols3> {
-        self.0.tls_config.alpn_protocols = vec![b"h2".to_vec()];
+        #[cfg(feature = "http1")]
+        let alpn_protocols = vec![b"h2".to_vec(), b"http/1.1".to_vec()];
+        #[cfg(not(feature = "http1"))]
+        let alpn_protocols = vec![b"h2".to_vec()];
+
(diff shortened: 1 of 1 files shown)


src/connector/builder.rs
```diff
@@ -174,16 +174,22 @@ impl ConnectorBuilder<WantsProtocols1> {
         })
     }
 
-    /// Enable all HTTP versions
+    /// Enable all HTTP versions built into this library
```
djc, line 177: Let's make this more explicit, like `(enabled with Cargo features)`?

src/connector/builder.rs
```diff
@@ -174,16 +174,22 @@ impl ConnectorBuilder<WantsProtocols1> {
...
     ///
-    /// For now, this enables both HTTP 1 and 2. In the future, other supported versions
-    /// will be enabled as well.
-    #[cfg(all(feature = "http1", feature = "http2"))]
+    /// For now, this could enable both HTTP 1 and 2, depending on active features.
+    /// In the future, other supported versions will be enabled as well.
+    #[cfg(feature = "http2")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "http2")))]
```
djc, line 182: If we're going to do this, let's get rid of the `http2` guard? And make it work for all of the possible combinations?

djc (commented):
I think this probably makes sense (and makes the interface easier to understand). @cpu any thoughts?

src/connector/builder.rs
cpu: Agreed, that seems like the right direction

src/connector/builder.rs
```diff
@@ -335,5 +341,14 @@ mod tests {
             &connector.tls_config.alpn_protocols,
             &[b"h2".to_vec(), b"http/1.1".to_vec()]
         );
+        let connector = super::ConnectorBuilder::new()
```
cpu, line 344: I think this test is only checking the result of using `enable_all_versions` with both the http1 and http2 feature flags enabled based on the guard on `test_alpn`. Ideally we could assert the correct state for more combinations than just this case.

src/connector/builder.rs
TimLuq: @djc

The problem with not having the `http2` guard is that the return type would change (since `WantsProtocols3` wont exist). So if any library enables the feature it might cause other libraries to not compile if they expected `WantsProtocols2`. Since enabling an extra feature in a deep dependency doesn't usually warrant a major version release, things in the wild could spontaneously break if one single library enables it when one other doesn't expect it.

Making the feature combination non-breaking would require wrapping the actual inner type `WantsProtocolsN` (based on the highest enabled protocol feature) in a `WantsProtocolsAll` singleton struct and always returning that. That would be a breaking change for this library, however, which is why I didn't want to do it in this fix.

If a version of this fix gets merged I could create a new PR with the above solution with a new breaking return type.

src/connector/builder.rs
TimLuq: > Ideally we could assert the correct state for more combinations than just this case.

Yeah, I realized this too. I locally created an additional test case for only `http2` while I was writing the PR description. I suspected that I'd get feedback on the doc line so I didn't want to muddy the waters with an additional small commit. But I'll push it now.

src/connector/builder.rs
djc: That seems reasonable, I'd be willing to merge this in order to get this improvement out provided you submit an additional PR to do the semver-breaking fix.

djc (approved):

src/connector/builder.rs
```diff
@@ -174,7 +174,7 @@ impl ConnectorBuilder<WantsProtocols1> {
         })
     }
 
-    /// Enable all HTTP versions built into this library
+    /// Enable all HTTP versions built into this library (enabled with Cargo features)
```
cpu, line 177: small nit: I think this change would ideally be squished into the first commit since its unrelated to the test coverage.

cpu (approved):
Thanks :-)

djc:
Thanks!

src/connector/builder.rs
cpu: We'll just squash merge this nit away.

cpu:
+1, thank you

PR #228, opened 2023-10-31
Remove undocumented panic in with_native_roots

### LINKED ISSUE #187: panics "no CA certificates found"
Hi,

I have a project which have indirect dependency to `hyper-rustls`. During runtime, the binary panics:
```
thread '<unnamed>' panicked at 'no CA certificates found', /home/pierre/.cargo/registry/src/github.com-1ecc6299db9ec82
3/hyper-rustls-0.22.1/src/connector.rs:45:13
```

The version is `0.22.1`.
Note that the crash happens on Android device.

kayabaNerve:
Breaking API change as required to implement this safety.

Somewhat resolves #187.

### DIFF
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -60,7 +60,7 @@ async fn run_client() -> io::Result<()> {
         // Default TLS client config with native roots
         None => rustls::ClientConfig::builder()
             .with_safe_defaults()
-            .with_native_roots()
+            .with_native_roots()?
             .with_no_client_auth(),
     };
     // Prepare the HTTPS connector
--- a/src/config.rs
+++ b/src/config.rs
@@ -8,9 +8,14 @@ use rustls::{ClientConfig, ConfigBuilder, WantsVerifier};
 pub trait ConfigBuilderExt {
     /// This configures the platform's trusted certs, as implemented by
     /// rustls-native-certs
+    ///
+    /// This will return an error if no valid certs were found. In that case,
+    /// it's recommended to use `with_webpki_roots`.
     #[cfg(feature = "rustls-native-certs")]
     #[cfg_attr(docsrs, doc(cfg(feature = "rustls-native-certs")))]
-    fn with_native_roots(self) -> ConfigBuilder<ClientConfig, WantsTransparencyPolicyOrClientCert>;
+    fn with_native_roots(
+        self,
+    ) -> std::io::Result<ConfigBuilder<ClientConfig, WantsTransparencyPolicyOrClientCert>>;
 
     /// This configures the webpki roots, which are Mozilla's set of
     /// trusted roots as packaged by webpki-roots.
@@ -23,7 +28,9 @@ impl ConfigBuilderExt for ConfigBuilder<ClientConfig, WantsVerifier> {
     #[cfg(feature = "rustls-native-certs")]
     #[cfg_attr(docsrs, doc(cfg(feature = "rustls-native-certs")))]
     #[cfg_attr(not(feature = "logging"), allow(unused_variables))]
-    fn with_native_roots(self) -> ConfigBuilder<ClientConfig, WantsTransparencyPolicyOrClientCert> {
+    fn with_native_roots(
+        self,
+    ) -> std::io::Result<ConfigBuilder<ClientConfig, WantsTransparencyPolicyOrClientCert>> {
         let mut roots = rustls::RootCertStore::empty();
         let mut valid_count = 0;
         let mut invalid_count = 0;
@@ -45,9 +52,15 @@ impl ConfigBuilderExt for ConfigBuilder<ClientConfig, WantsVerifier> {
             valid_count,
             invalid_count
         );
-        assert!(!roots.is_empty(), "no CA certificates found");
+        if roots.is_empty() {
+            crate::log::debug!("no valid native root CA certificates found");
+            Err(std::io::Error::new(
+                std::io::ErrorKind::NotFound,
+                format!("no valid native root CA certificates found ({invalid_count} invalid)"),
+            ))?
+        }
 
-        self.with_root_certificates(roots)
+        Ok(self.with_root_certificates(roots))
     }
 
     #[cfg(feature = "webpki-roots")]
--- a/src/connector/builder.rs
+++ b/src/connector/builder.rs
@@ -59,13 +59,13 @@ impl ConnectorBuilder<WantsTlsConfig> {
     /// [with_safe_defaults]: rustls::ConfigBuilder::with_safe_defaults
     #[cfg(feature = "rustls-native-certs")]
     #[cfg_attr(docsrs, doc(cfg(feature = "rustls-native-certs")))]
-    pub fn with_native_roots(self) -> ConnectorBuilder<WantsSchemes> {
-        self.with_tls_config(
+    pub fn with_native_roots(self) -> std::io::Result<ConnectorBuilder<WantsSchemes>> {
+        Ok(self.with_tls_config(
             ClientConfig::builder()
                 .with_safe_defaults()
-                .with_native_roots()
+                .with_native_roots()?
                 .with_no_client_auth(),
-        )
+        ))
     }
 
     /// Shorthand for using rustls' [safe defaults][with_safe_defaults]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -14,6 +14,7 @@
 //! let url = ("https://hyper.rs").parse().unwrap();
 //! let https = hyper_rustls::HttpsConnectorBuilder::new()
 //!     .with_native_roots()
+//!     .expect("no native root CA certificates found")
 //!     .https_only()
 //!     .enable_http1()
 //!     .build();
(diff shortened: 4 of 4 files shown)


examples/client.rs
```diff
@@ -61,6 +61,7 @@ async fn run_client() -> io::Result<()> {
         None => rustls::ClientConfig::builder()
             .with_safe_defaults()
             .with_native_roots()
+            .expect("no CAs were present on the system")
```
cpu, line 64:
```suggestion
            .expect("no native root CAs were present on the system")
```

src/config.rs
```diff
@@ -10,7 +10,9 @@ pub trait ConfigBuilderExt {
     /// rustls-native-certs
     #[cfg(feature = "rustls-native-certs")]
     #[cfg_attr(docsrs, doc(cfg(feature = "rustls-native-certs")))]
-    fn with_native_roots(self) -> ConfigBuilder<ClientConfig, WantsTransparencyPolicyOrClientCert>;
+    fn with_native_roots(
```
cpu, line 13: I think the rustdoc comment for this trait fn should mention why it might return `None`, and to suggest using `with_webpki_roots` in that circumstance.

src/config.rs
```diff
@@ -45,9 +49,12 @@ impl ConfigBuilderExt for ConfigBuilder<ClientConfig, WantsVerifier> {
             valid_count,
             invalid_count
         );
-        assert!(!roots.is_empty(), "no CA certificates found");
+        if roots.is_empty() {
+            crate::log::debug!("no CA certificates found");
```
cpu, line 53:
```suggestion
            crate::log::debug!("no native root CA certificates found");
```

src/lib.rs
```diff
@@ -14,6 +14,7 @@
 //! let url = ("https://hyper.rs").parse().unwrap();
 //! let https = hyper_rustls::HttpsConnectorBuilder::new()
 //!     .with_native_roots()
+//!     .expect("no CAs were present on the system")
```
cpu, line 17:
```suggestion
//!     .expect("no native root CAs were present on the system")
```

src/lib.rs
```diff
@@ -59,6 +60,7 @@
 //! let key = rustls::PrivateKey(keys[0].clone());
 //! let https = hyper_rustls::HttpsConnectorBuilder::new()
 //!     .with_native_roots()
+//!     .expect("no CAs were present on the system")
```
cpu, line 63:
```suggestion
//!     .expect("no native root CAs were present on the system")
```

cpu (commented):
Hi, thanks for the PR. I have a few initial comments.

At a meta-level I'm also wondering whether we should augment this API surface with something feature-gated on webpki-roots that allows trying to find native trust anchors, but automatically falling back to bundled webpki-roots trust anchors when it comes up empty. I suspect that might be more user friendly, and matches the approach offered in rustls-platform-verifier (see some discussion in https://github.com/rustls/rustls-platform-verifier/issues/12).

The rustls-platform-verifier work was motivated by observation in the field that `openssl-probe` wasn't always reliable, and that's the dependency underpinning rustls-native-roots on non-MacOS unix systems. What platform you're seeing this panic on? There were similar requests for more information in #187 that nobody has provided to date.

kayabaNerve:
> but automatically falling back to bundled webpki-roots trust anchors when it comes up empty. I suspect that might be more user friendly, and matches the approach offered in rustls-platform-verifier (see some discussion in [rustls/rustls-platform-verifier#12](https://github.com/rustls/rustls-platform-verifier/issues/12)).

While I'd support this, I believe that should be an additional, recommended function. Not the replacement behavior for this function.

I observed this on a Linux system without ca-certificates installed. Specifically, a GH CI runner I uninstalled most packages from. Then, I also encountered it on a Debian slim Docker image.

I'm fine updating the documentation :+1:

kayabaNerve:
Done

src/config.rs
```diff
@@ -23,7 +28,9 @@ impl ConfigBuilderExt for ConfigBuilder<ClientConfig, WantsVerifier> {
     #[cfg(feature = "rustls-native-certs")]
     #[cfg_attr(docsrs, doc(cfg(feature = "rustls-native-certs")))]
     #[cfg_attr(not(feature = "logging"), allow(unused_variables))]
-    fn with_native_roots(self) -> ConfigBuilder<ClientConfig, WantsTransparencyPolicyOrClientCert> {
+    fn with_native_roots(
+        self,
+    ) -> Option<ConfigBuilder<ClientConfig, WantsTransparencyPolicyOrClientCert>> {
```
ctz, line 33: my feeling is that this should be returning a `Result` with a `Box<dyn std::error::Error>` or `std::io::Error`

then the `.expect("could not load platform certs")` line below can become `?` to avoid a reachable panic there.

src/config.rs
```diff
@@ -45,9 +52,12 @@ impl ConfigBuilderExt for ConfigBuilder<ClientConfig, WantsVerifier> {
             valid_count,
             invalid_count
         );
-        assert!(!roots.is_empty(), "no CA certificates found");
+        if roots.is_empty() {
+            crate::log::debug!("no native root CA certificates found");
+            None?
```
ctz, line 57: and this can return something more useful, -- i suggest including the logged message, and the `valid_count` and `invalid_count` diagnostics too (i appreciate there's logging going on, but it's less immediate than returning the same stuff in a `Result::Err`, even if it gets smushed into a String along the way)

src/connector/builder.rs
```diff
@@ -59,12 +59,14 @@ impl ConnectorBuilder<WantsTlsConfig> {
     /// [with_safe_defaults]: rustls::ConfigBuilder::with_safe_defaults
     #[cfg(feature = "rustls-native-certs")]
     #[cfg_attr(docsrs, doc(cfg(feature = "rustls-native-certs")))]
-    pub fn with_native_roots(self) -> ConnectorBuilder<WantsSchemes> {
-        self.with_tls_config(
-            ClientConfig::builder()
-                .with_safe_defaults()
-                .with_native_roots()
-                .with_no_client_auth(),
+    pub fn with_native_roots(self) -> Option<ConnectorBuilder<WantsSchemes>> {
```
ctz, line 62: ditto, `Result`

djc:
Usually we would postpone builder API errors until the end of the builder chain, which this would subvert.

kayabaNerve:
While I'm not against standing in the way in convention with PR, I'd note that policy makes it more difficult to identify which step in the process failed as you get one Result for all steps, not one Result for each step which may fail. While the error itself may reasonably allowing discovery of the issue, the proposed path here is if `with_native_roots` returns None/error, callers fallback to `with_webpki_roots`. They'd need a way to introspect the final Result to check if it's due to this specific issue in order to safely execute the suggested fallback (unless they simply blindly try again with `with_webpki_roots`).

Since the suggested error to replace this with was `std:io::Error`, not `hyper_rustls::config::BuildError`, this seems infeasible.

Also, while I do agree the error would be more descriptive, I think None to signify there were no valid roots on the system is fine. Regardless, I'm happy to move it to an error, so long as the policy on *where to put the error* and the suggested mediation path is resolved.

kayabaNerve:
@ctz Resolved your comments :)

src/config.rs
```diff
@@ -8,9 +8,14 @@ use rustls::{ClientConfig, ConfigBuilder, WantsVerifier};
 pub trait ConfigBuilderExt {
     /// This configures the platform's trusted certs, as implemented by
     /// rustls-native-certs
+    ///
+    /// This will return None if no valid certs were found. In that case,
```
cpu, line 12: I think this comment fell out-of-date with the impl after the change to use `Result`.

kayabaNerve: It did, sorry. Fixed!

kayabaNerve:
Also, the lib.rs example is failing as its function doesn't return an error, so that does have to be an expect...

cpu (approved):
Thank you.

I will circle back with the idea of making a new combined "try native then webpki-roots" method later on (unless you want to tackle that in a follow-up yourself).

kayabaNerve:
*Personally*, I think that's trivial enough it doesn't have to be in hyper-rustls, but it's hard to argue against libs adding extra helpers to go the extra mile in friendliness...

kayabaNerve:
Anything else needed for this to be merged?

cpu:
> Anything else needed for this to be merged?

Looks like there's a CI failure from another examples in `lib.rs` that need to `expect`. Could you fix that up @kayabaNerve? It would also be helpful if you could squash your commit history.

kayabaNerve:
Oh, sorry. I have no idea how I missed that. Will correct that now.

Would love to get this included with whatever release updates to 1.0 :)

kayabaNerve:
Pushed a new single commit.

kayabaNerve:
CI passed :D

ctz (approved):
this looks good to me

cpu (approved):

djc (approved):

cpu:
Thanks kayabaNerve.

kayabaNerve:
Happy to.

PR #232, opened 2023-11-18
Update hyper to 1.x and integrate with hyper-util

Gelbpunkt:
Since hyper 1.x leaves control of the accepting process to the user ([explanation](https://github.com/rustls/hyper-rustls/pull/232#issuecomment-1879272874)), I removed the acceptor API and server examples.

The tokio feature in hyper-util is required for the IO adapter traits, so the need for a tokio-runtime feature is gone (it was already a hard dependency before).


Gelbpunkt:
I marked this as ready for review but since I still didn't hear back regarding the future of the acceptor API, it isn't really "ready". If it were up to me, I'd completely remove the acceptor API, but I'm not the one to make that decision so I'd really prefer hearing from the maintainers.

This PR is blocking a lot of upgrades to hyper 1.0 in the ecosystem, I'd really appreciate hearing back so this PR can be reviewed properly.

cpu:
Thanks for the PR. This is a substantial update and I'm not as familiar with the guts of hyper-rustls as I am with other parts of the Rustls ecosystem. Spinning up on the Hyper 1.0 API changes and the work in this branch will take time that I haven't been able to find yet. Your patience is appreciated. 

> If it were up to me, I'd completely remove the acceptor API

Have you looked at any downstream consumers of this crate to see if they are using the acceptor API, and what they would need to do in place of it were it removed?

Gelbpunkt:
> Have you looked at any downstream consumers of this crate to see if they are using the acceptor API, and what they would need to do in place of it were it removed?
Thing is, the acceptor trait was removed from hyper entirely. There is no more direct purpose in having something that previously implemented the trait living in hyper-rustls anymore. hyper pre-1.0 had a high-level server API that would use the trait, this high-level API was removed without replacement. In hyper 1.0, users have to create the TcpListener themselves and drive the HTTP connection on the TcpStream themselves.

Let me demonstrate what this looks like without TLS.

```rs
// hyper 0.14
use std::convert::Infallible;
use std::net::SocketAddr;
use hyper::{Body, Request, Response, Server};
use hyper::service::{make_service_fn, service_fn};

async fn handle(_req: Request<Body>) -> Result<Response<Body>, Infallible> {
    Ok(Response::new(Body::from("Hello World")))
}

#[tokio::main]
async fn main() {
    // Construct our SocketAddr to listen on...
    let addr = SocketAddr::from(([127, 0, 0, 1], 3000));

    // And a MakeService to handle each connection...
    let make_service = make_service_fn(|_conn| async {
        Ok::<_, Infallible>(service_fn(handle))
    });

    // Then bind and serve...
    let server = Server::bind(&addr).serve(make_service);

    // And run forever...
    if let Err(e) = server.await {
        eprintln!("server error: {}", e);
    }
}
```

And here's the hyper 1.x equivalent:
```rs
// Hyper 1.x
use http_body_util::Full;
use hyper::body::{Bytes, Incoming};
use hyper::service::service_fn;
use hyper::{Request, Response};
use hyper_util::rt::{TokioExecutor, TokioIo};
use hyper_util::server::conn::auto;
use std::convert::Infallible;
use std::net::SocketAddr;
use tokio::net::TcpListener;

async fn handle(_req: Request<Incoming>) -> Result<Response<Full<Bytes>>, Infallible> {
    Ok(Response::new(Full::from("Hello World")))
}

#[tokio::main]
async fn main() {
    // Construct our SocketAddr to listen on...
    let addr = SocketAddr::from(([127, 0, 0, 1], 3000));

    // Bind to the address
    let listener = TcpListener::bind(addr).await.unwrap();

    // Serve incoming connections
    loop {
        let (stream, _) = listener.accept().await.unwrap();

        // Note that we have to drive the connections ourselves and also take care of the accepting process
        tokio::spawn(async move {
            auto::Builder::new(TokioExecutor::new())
                .serve_connection(TokioIo::new(stream), service_fn(handle))
                .await
                .unwrap();
        });
    }
}
```

You'll notice that the accepting of connections is now no longer up to hyper's now-removed `Server`, but rather the user. That is also why the `Acceptor` trait is removed: There is no more need for it. It was needed to customize the `accept` process of the `Server`, but now the user is in charge of this.

All that the hyper-rustls `TlsAcceptor` did was wrap the I/O received from hyper after calling `accept` in a `TlsStream` ([relevant line](https://docs.rs/hyper-rustls/latest/src/hyper_rustls/acceptor.rs.html#53)). That is something that users can just do themselves now after they have accepted the connection. There's nothing hyper-specific about this anymore, therefore it doesn't need to reside in hyper-rustls anymore in my opinion. 

I hope that somewhat explains why I think this is pointless now. Any "new" implementation of `TlsAcceptor` like the one I currently provide in this PR is quite literally a fancy wrapper over `TlsStream::new` coupled with a builder for `Arc<ServerConfig>`. I don't see why that should exist anymore.

kurtbuilds:
@Gelbpunkt Want to say thank you for putting the PR together. Wanted to chime in and confirm, I've got tons of code that is blocked on upgrading http=1.0 (and axum=0.7, etc) because this isn't merged.

src/connector.rs
```diff
@@ -4,12 +4,12 @@ use std::sync::Arc;
...
-use hyper::client::connect::Connection;
-use hyper::service::Service;
-use hyper::Uri;
+use http::Uri;
+use hyper::rt::{Read, Write};
+use hyper_util::rt::TokioIo;
 use pki_types::ServerName;
-use tokio::io::{AsyncRead, AsyncWrite};
 use tokio_rustls::TlsConnector;
+use tower_service::Service;
```
cpu, line 12: The Hyper 1.0 [upgrade guide](https://hyper.rs/guides/1/upgrading/) says:
> Previously, hyper depended on tower for the Service trait. Because tower is not yet 1.0, hyper could not publicly depend on it. So, it’s Service trait and the helper function service_fn are defined in hyper::service.

I'm not clear on why we need to use a direct dependency on `tower_service` and its `Service` trait here instead of the [hyper::service::Service](https://docs.rs/hyper/latest/hyper/service/trait.Service.html) trait. Can you elaborate on this part of the changeset?

examples/server.rs
```diff
@@ -35,42 +41,55 @@ async fn run_server() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
         Some(ref p) => p.to_owned(),
         None => "1337".to_owned(),
     };
-    let addr = format!("127.0.0.1:{}", port).parse()?;
+    let addr: SocketAddr = format!("127.0.0.1:{}", port).parse()?;
```
cpu, line 44: nit: I think we generally prefer to use a turbofish for this sort of thing.

examples/server.rs
```diff
@@ -35,42 +41,55 @@ async fn run_server() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
 
     // Load public certificate.
     let certs = load_certs("examples/sample.pem")?;
     // Load private key.
     let key = load_private_key("examples/sample.rsa")?;
-    // Build TLS configuration.
+
+    println!("Starting to serve on https://{}.", addr);
```
cpu, line 51: nit: I'd drop the trailing `.` here:
```suggestion
    println!("Starting to serve on https://{}", addr);
```

cpu:
> Let me demonstrate what a hyper server looks like without TLS in 0.14 and 1.x so I can better explain the changes to the accept process.

Thank you, this was helpful :+1: 

> I don't see why that should exist anymore.

I'm convinced. I think you should go ahead and remove it from this branch (and update the PR description).

cpu (commented):
some questions/small comments from a first pass. Thanks!

examples/server.rs
djc: This is a pretty roundabout way of doing it, I'd suggest writing it as `SocketAddr::from(([127, 0, 0, 1], port))`.

src/connector.rs
Gelbpunkt: Because hyper-util has different trait bounds. See https://docs.rs/hyper-util/latest/hyper_util/client/legacy/connect/trait.Connect.html, it requires the `tower_service` `Service<Uri>` trait to be implemented for the connector and not hyper's.

examples/server.rs
Gelbpunkt: I'll remove this example since it was previously being used to showcase usage of the `Acceptor`, which I'm removing now.

examples/server.rs
Gelbpunkt: See my previous comment about removing the server example

src/connector.rs
```diff
@@ -38,7 +38,7 @@ impl<T> HttpsConnector<T> {
 impl<T> Service<Uri> for HttpsConnector<T>
 where
     T: Service<Uri>,
-    T::Response: Connection + AsyncRead + AsyncWrite + Send + Unpin + 'static,
```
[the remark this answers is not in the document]
Gelbpunkt: ... that is a good question, I wonder how this even compiles. [Connect requires the `Service` response to implement Connection](https://docs.rs/hyper-util/latest/hyper_util/client/legacy/connect/trait.Connect.html), which the `MaybeHttpsStream` it only does if `T` implements `Connection`... I assume in reality it would error at compile time if you were to use it outside of `TokioIo<TcpStream>`. I'll add the bound to be sure

examples/server.rs
Arnavion: I think it would be nice if you adapted it to the new API you showed in https://github.com/rustls/hyper-rustls/pull/232#issuecomment-1879272874 instead of deleting it, since probably a lot of users will want to know how to do it, especially if they're existing users of the current rustls version and looking for the upgrade path.

src/connector.rs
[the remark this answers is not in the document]
Gelbpunkt: Ah I think I got it: Since we require `T: Service<Uri>`, `T::Response` is guaranteed to implement the traits based on the bounds in hyper-util

Gelbpunkt:
One issue that arises with removing the server example is that there's no longer a test that runs the client example with custom certificates. I assume the best way forward is to inline a server that does this into the tests? I'm thinking that it seems a bit weird that the server and the client were testing "each other" previously, since a misimplementation might go unnoticed if it's in hyper-rustls. Wouldn't it be better to test the client's custom certificate with a completely different server implementation (similar to how the server was tested with curl)?

src/connector.rs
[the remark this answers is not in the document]
cpu: Thanks! Resolving based on this thread & 6fca9a1cd5eca0de313522c244895727be38fb6d

cpu:
Looking good :-)

> One issue that arises with removing the server example is that there's no longer a test that runs the client example with custom certificates. 

Could this issue and [@Arnavion's feedback](https://github.com/rustls/hyper-rustls/pull/232#discussion_r1445047784) be addressed at the same time by adding an example server that doesn't use the now-removed `Acceptor` abstraction, but runs an HTTPS server looping over accepting connections from a `TcpListener`? That is, keep the example, but remove only the `Acceptor` specific parts.

>  I'm thinking that it seems a bit weird that the server and the client were testing "each other" previously, since a misimplementation might go unnoticed if it's in hyper-rustls. Wouldn't it be better to test the client's custom certificate with a completely different server implementation (similar to how the server was tested with curl)?

I think it's fair to think about this as follow-up work and not block the hyper update on improving the tests like this.

> rustls / Check semver compatibility (pull_request) Failing after 24s

If you want to update the `Cargo.toml` version to 0.26.0 this should resolve this CI task failing. Squashing your commit history would also be appreciated. Ideally this branch would be two commits: all of the existing changes as one Hyper update commit, and then the `Cargo.toml` version update as a second commit.

examples/server.rs
Gelbpunkt: I've added such an example back and addressed the comments left on this old one there.

Gelbpunkt:
> One note, tower-service 0.3 depends on http 0.2. This fact doesn't block this work at all, but it does indicate that, to drop the http 0.2 dependency, we need either an updated hyper-util that depends on hyper::Service rather than tower-service::Service (I'm guessing this is preferable, but I don't know if it's feasible), or an updated tower-service that depends on http=1.0.

That is a dev-dependency and therefore not relevant for us.

cpu (approved):
Looks good to me :rocket:

src/stream.rs
```diff
@@ -15,19 +16,19 @@ pub enum MaybeHttpsStream<T> {
     /// A stream over plain text.
     Http(T),
     /// A stream protected with TLS.
-    Https(TlsStream<T>),
+    Https(TokioIo<TlsStream<TokioIo<T>>>),
```
djc, line 19: This `TokioIo<TlsStream<TokioIo<T>>>` business seems pretty ugly to me. Is there some reasonable way we can do this with fewer layers of stuff? I guess we're contrained by what tokio_rustls implements for `TlsStream`?

src/stream.rs
```diff
@@ -15,19 +16,19 @@ pub enum MaybeHttpsStream<T> {
...
 }
 
-impl<T: AsyncRead + AsyncWrite + Connection + Unpin> Connection for MaybeHttpsStream<T> {
+impl<T: Read + Write + Connection + Unpin> Connection for MaybeHttpsStream<T> {
```
djc, line 22: Nit: I'd prefer to use `rt::Read + rt::Write` since the trait names are also used in `std` and I'd like to clearly disambiguate from those.

(Same for the other uses below.)

djc (commented):
Thanks for working on this and sorry for taking so long to review this in more depth.

I think this is looking pretty good but still have few comments.

src/stream.rs
Gelbpunkt: So, our `T` is required to implement the hyper I/O traits so we can delegate them when we implement the I/O traits  for `MaybeHttpsStream`. tokio-rustls' `TlsStream` requires the tokio I/O traits, so we use a `TokioIo` to bridge *from* hyper I/O traits *to* tokio I/O traits. And because the `TlsStream` only implements the tokio traits, we then use another `TokioIo` to bridge to the hyper traits again.

Gelbpunkt: In fact that is how hyper-tls does it as well: https://docs.rs/hyper-tls/latest/src/hyper_tls/stream.rs.html#20
Not that pretty but that's the only way to make this work to my knowledge

djc: Alright, I guess it's fine for now.

src/stream.rs
Gelbpunkt: Done and squashed.

src/connector.rs
```diff
@@ -38,7 +39,7 @@ impl<T> HttpsConnector<T> {
...
+    T::Response: Connection + Read + Write + Send + Unpin + 'static,
```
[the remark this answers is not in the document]
Gelbpunkt: Sorry for missing that, updated.

examples/server.rs
```diff
@@ -1,20 +1,27 @@
...
-use hyper::{Body, Method, Request, Response, Server, StatusCode};
-use hyper_rustls::TlsAcceptor;
+use http::{Method, Request, Response, StatusCode};
+use http_body_util::{BodyExt, Full};
+use hyper::{
+    body::{Bytes, Incoming},
+    service::service_fn,
+};
+use hyper_util::rt::TokioIo;
+use hyper_util::{rt::TokioExecutor, server::conn::auto::Builder};
```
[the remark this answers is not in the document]
Gelbpunkt: Done!

cpu (approved):

cpu:
Thanks again @Gelbpunkt :bow:

cpu:
Tracking a release with this work in https://github.com/rustls/hyper-rustls/issues/250

PR #233, opened 2023-11-24
Update to rustls alpha releases

djc:


examples/server.rs
```diff
@@ -80,34 +81,26 @@ async fn echo(req: Request<Body>) -> Result<Response<Body>, hyper::Error> {
...
-        .map_err(|_| error("failed to load private key".into()))?;
-    if keys.len() != 1 {
-        return Err(error("expected a single private key".into()));
-    }
-
-    Ok(rustls::PrivateKey(keys[0].clone()))
+    let result = rustls_pemfile::rsa_private_keys(&mut reader)
+        .next()
+        .unwrap();
+    result.map(|key| key.into())
```
cpu, line 105: nit: seems tidier to do this all in one go without the `result` binding:

```suggestion
    rustls_pemfile::rsa_private_keys(&mut reader)
        .next()
        .unwrap()
        .map(|key| key.into())
```

djc: Cleaned this up.

djc:
Bumped this to the next batch of alphas, want to take another look?

PR #235, opened 2023-11-30
Update to rustls 0.22.0 alpha.6

tottoto:
Updates to `rustls` 0.22.0-alpha.6, fixes a feature requirement, and adds a ci job to check each feature.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -145,3 +145,17 @@ jobs:
         with:
           components: clippy
       - run: cargo clippy --all-features -- --deny warnings
+
+  features:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - name: Install rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - name: Install cargo-hack
+        uses: taiki-e/install-action@cargo-hack
+      - name: Check feature powerset
+        run: cargo hack --no-dev-deps check --feature-powerset --depth 2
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -145,3 +145,17 @@ jobs:
...
+      - name: Checkout sources
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - name: Install rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - name: Install cargo-hack
+        uses: taiki-e/install-action@cargo-hack
+      - name: Check each features
+        run: cargo hack --no-dev-deps check --each-feature
```
cpu, line 161: WDYT about using `--feature-powerset ` instead of `--each-feature`? Would it inflate CI time too substantially?

cpu (approved):

.github/workflows/build.yml
tottoto: Updates to use `--feature-powerset` to test the CI time.

tottoto: In this test, the CI time increased from `32s` (https://github.com/tottoto/hyper-rustls/actions/runs/7050201648/job/19190355371) to `44s` (https://github.com/rustls/hyper-rustls/actions/runs/7050273761/job/19190655897). I think it is within an acceptable range as the impact on the overall CI is small.

cpu: Agreed, that seems totally reasonable to me. Would you mind squashing the commit to use the powerset into the commit adding the job?

tottoto: Squashed the two commits.

cpu: Thanks!

djc:
Thanks!

PR #245, opened 2023-12-11
Style tweaks

djc:
Clean up code style in hyper-rustls before refactoring to support hyper 1, bringing things in line with rustls [code style](https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md).

### DIFF
--- a/src/connector.rs
+++ b/src/connector.rs
@@ -102,20 +102,19 @@ where
                 return Box::pin(async move { Err(Box::new(err).into()) });
             }
         };
-        let connecting_future = self.http.call(dst);
 
-        let f = async move {
+        let connecting_future = self.http.call(dst);
+        Box::pin(async move {
             let tcp = connecting_future
                 .await
                 .map_err(Into::into)?;
-            let connector = TlsConnector::from(cfg);
-            let tls = connector
-                .connect(hostname, tcp)
-                .await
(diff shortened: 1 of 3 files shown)


src/connector.rs
```diff
@@ -4,7 +4,9 @@ use std::sync::Arc;
 use std::task::{Context, Poll};
 use std::{fmt, io};
 
-use hyper::{client::connect::Connection, service::Service, Uri};
+use hyper::client::connect::Connection;
```
ctz, line 7: similar change in acceptor.rs?

```
diff --git a/src/acceptor.rs b/src/acceptor.rs
index 037b64e..02e72d0 100644
--- a/src/acceptor.rs
+++ b/src/acceptor.rs
@@ -5,10 +5,8 @@ use std::pin::Pin;
 use std::sync::Arc;
 
 use futures_util::ready;
-use hyper::server::{
-    accept::Accept,
-    conn::{AddrIncoming, AddrStream},
-};
+use hyper::server::accept::Accept;
+use hyper::server::conn::{AddrIncoming, AddrStream};
```

ctz (approved):

src/connector.rs
djc: Good call!

cpu (commented):
Very nice, thank you

PR #253, opened 2024-01-24
Add built-in support for rustls-platform-verifier

djc:
r? @ComplexSpaces

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,4 +1,11 @@
-#[cfg(any(feature = "rustls-native-certs", feature = "webpki-roots"))]
+#[cfg(feature = "rustls-platform-verifier")]
+use std::sync::Arc;
+
+#[cfg(any(
+    feature = "rustls-platform-verifier",
+    feature = "rustls-native-certs",
+    feature = "webpki-roots"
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -7,6 +14,10 @@ use rustls::{ClientConfig, ConfigBuilder, WantsVerifier};
 /// This adds methods (gated by crate features) for easily configuring
 /// TLS server roots a rustls ClientConfig will trust.
 pub trait ConfigBuilderExt {
+    /// Use the platform's native verifier to verify server certificates.
```
cpu, line 17: WDYT about putting a link to the platform verifier crate here? I think it might be non-obvious what platforms are supported, or what the various per-platform capabilities are. Duplicating that info here will be problematic, but we could at least give a pointer to the platform verifier repo/docs.

cpu (approved):

src/config.rs
djc: Done.

complexspaces (commented):
It seems like I'm late to reviewing this, but this looks good to me as well 👍

PR #255, opened 2024-02-19
Cargo.toml: add aws-lc-rs feature as crypto backend

BiagioFesta:


cpu (commented):
It seems reasonable to expose this so that the default ring feature can be disabled and aws-lc-rs enabled :+1: 

> rustls / Build+test (windows-latest, stable) (pull_request) Failing after 3m
> ..
>   -- Didn't find assembler

I think CI will need adjusting to [add some extra build deps](https://github.com/rustls/rustls/blob/fdf71f8ed046d1e8680aa02858b22d69c36b41fb/.github/workflows/build.yml#L42-L48) for Windows.

cpu (approved):
Thanks

djc (approved):

PR #259, opened 2024-03-15
feat: add a builder() method to HttpsConnector

Fishrock123:
It's become very idiomatic to have the builder api be accessed this way.

### DIFF
--- a/src/connector.rs
+++ b/src/connector.rs
@@ -28,6 +28,13 @@ pub struct HttpsConnector<T> {
 }
 
 impl<T> HttpsConnector<T> {
+    /// Creates a [`crate::HttpsConnectorBuilder`] to configure a `HttpsConnector`.
+    ///
+    /// This is the same as [`crate::HttpsConnectorBuilder::new()`].
+    pub fn builder() -> builder::ConnectorBuilder<builder::WantsTlsConfig> {
+        builder::ConnectorBuilder::new()
+    }
+
     /// Force the use of HTTPS when connecting.
     ///
     /// If a URL is not `https` when connecting, an error is returned.


Fishrock123:
Not quite sure how to make rustdoc link those references correctly? 😕

djc:
See https://github.com/rustls/hyper-rustls/pull/260#issuecomment-2003318005.

Fishrock123:
This PR could just as well be completely unrelated from the other two.

This is not _necessary_ but is a common API idom for builder APIs in Rust.

djc (approved):
Fair point! I think this is a good change.

Tiny nit: as a constructor of sorts (even if indirectly), I would prefer we move `builder()` before `enforce_https()`.

cpu (approved):
Thanks, this looks good to me modulo the CI docs failure getting fixed.

> Not quite sure how to make rustdoc link those references correctly?

I think this diff will do the trick:

```diff
diff --git a/src/connector.rs b/src/connector.rs
index 103dede..5b975e0 100644
--- a/src/connector.rs
+++ b/src/connector.rs
@@ -35,9 +35,9 @@ impl<T> HttpsConnector<T> {
         self.force_https = true;
     }

-    /// Creates a [`builder::HttpsConnectorBuilder`] to configure a `HttpsConnector`.
+    /// Creates a [`crate::HttpsConnectorBuilder`] to configure a `HttpsConnector`.
     ///
-    /// This is the same as [`builder::HttpsConnectorBuilder::new()`].
+    /// This is the same as [`crate::HttpsConnectorBuilder::new()`].
     pub fn builder() -> builder::ConnectorBuilder<builder::WantsTlsConfig> {
         builder::ConnectorBuilder::new()
     }
```

PR #260, opened 2024-03-15
feat: wrap_connector from builder WantsProtocols1

Fishrock123:
There is a use case for wrapping libraries where they may not bring in either `http` or `http2` features, but still want to set advanced configuration but with their own lower-level connector.

This was discovered when trying to expose `builder::with_server_name()` in `reqwest`, which seems impossible right now without enabling extra features.

### DIFF
--- a/src/connector/builder.rs
+++ b/src/connector/builder.rs
@@ -228,6 +228,15 @@ impl ConnectorBuilder<WantsProtocols1> {
         })
     }
 
+    /// This wraps an arbitrary low-level connector into an [`HttpsConnector`]
+    pub fn wrap_connector<H>(self, conn: H) -> HttpsConnector<H> {
+        // HTTP1-only, alpn_protocols stays empty
+        // HttpConnector doesn't have a way to say http1-only;
+        // its connection pool may still support HTTP2
+        // though it won't be used
+        self.0.wrap_connector(conn)
+    }
+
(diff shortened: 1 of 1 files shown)


djc:
This is part of a trio of smallish PRs:

- #259 
- #260 
- #261 

Just to make sure we're not losing the forest for the trees, what exactly is the problem that you're trying to solve here? Adding a bunch of API surface like this has me wary that this might not be the best way to address the underlying needs.

Fishrock123:
These PRs are separate because they are in effect all completely unrelated, and could just as well be discovered independently.

This one is the most important. Reqwest may not be able to use the builder API without this change. (I am not a request author, so I can't speak to if they would accept the PR with added, unecessary, `hyper-rustls` features.)

Put more broadly, if a wrapping client wants to implement the `HttpConnector` itself and not pull in unnecessary code, this is required, because otherwise you cannot advance the builder's static state far enough.

cpu:
> This one is the most important. Reqwest may not be able to use the builder API without this change. (I am not a request author, so I can't speak to if they would accept the PR with added, unecessary, hyper-rustls features.)

If this change is motivated by a need in Reqwest, could you open a draft PR sketching out that side of the equation that takes a git dep on this branch? I think if my understanding is correct we would want to get stronger signal from the consuming projects before adding more API surface here.

PR #261, opened 2024-03-15
feat: builder with shared configuration

Fishrock123:
Sometimes one might want to share the config to the builder api like one can to the connector api itself, allowing some optimization.

Since the config eventually get's `Arc`-d anyways there can be no disadvantage to this.


src/connector/builder.rs
```diff
@@ -44,6 +46,27 @@ impl ConnectorBuilder<WantsTlsConfig> {
     /// [`enable_http2`](ConnectorBuilder::enable_http2)) before the
     /// connector is built.
     pub fn with_tls_config(self, config: ClientConfig) -> ConnectorBuilder<WantsSchemes> {
+        assert!(
```
djc, line 49: I'd propose we instead take `impl Into<Arc<ClientConfig>>` for `with_tls_config()`, which can cover both cases?

djc:
If htis is for reqwest, not sure this still make sense if it wants to just set its own fully-configured `ClientConfig`? This seems like a pretty niche use case and I'm not sure this is the best way to cover it.

src/connector/builder.rs
Fishrock123: That would probably be ideal; I was trying to avoid making a breaking change but I could do that if it was preferred.

Fishrock123:
It wants to set it's own fully configured `ClientConfig` but already holds an `Arc`-shared reference, hence this PR.

I can imagine other clients may like this as well.

src/connector/builder.rs
Fishrock123: Ahhhh I see what you are saying. I did not realize that the stdlib implemented [`From<T> for Arc<T>`](https://doc.rust-lang.org/std/sync/struct.Arc.html#impl-From%3CT%3E-for-Arc%3CT%3E).

Fishrock123:
Updated to have `pub fn with_tls_config(self, config: impl Into<Arc<ClientConfig>>) ...`

djc:
Why does it hold an `Arc`-shared reference? I don't like the use of `Arc::make_mut()` here -- it seems weird to me to want to use the `HttpConnectorBuilder` for a `ClientConfig` that is already "shrinkwrapped" in `Arc`.

Fishrock123:
> Why does it hold an `Arc`-shared reference? 

I'll have to spend more time in the codebase to answer that in an exact way - I don't know offhand, if I had to guess from past experience with Surf then connection pooling seems like a likely reason.

I could change this to [`Arc::get_mut`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.get_mut) if you like, which could possibly be a better way of doing things?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "rust-analyzer.cargo.features": "all"
+}
```
cpu, line 3: Was this part of the diff intentional? I think we'd prefer to see a `.vscode` addition to `.gitignore` in a stand-alone commit and this file backed out. There isn't an established precedent for committing IDE settings to the Rustls repos AFAIK.

Fishrock123: Unintentional, but it might take me a bit to get to it now I've got some life stuff occuring

cpu: No worries, best of luck.

PR #263, opened 2024-03-22
Update rustls to 0.23

howardjohn:


Cargo.toml
```diff
@@ -18,17 +18,17 @@ log = { version = "0.4.4", optional = true }
 pki-types = { package = "rustls-pki-types", version = "1" }
 rustls-native-certs = { version = "0.7", optional = true }
 rustls-platform-verifier = { version = "0.2", optional = true }
```
cpu, line 20: I think this version will also need updating to a not-yet-released 0.3 that includes https://github.com/rustls/rustls-platform-verifier/pull/70

(_I'm also surprised you can build/pass the test suite without any code changes :thinking: I would have expected changes would be needed for the process wide provider_)

howardjohn: Yep just figured that out when I tested with the platform verifier feature on. Bummer, will pause on that.

For provider, my understanding is this is (implicitly) relying on `CryptoProvider::get_default_or_install_from_crate_features()` for the examples/tests

cpu: > Yep just figured that out when I tested with the platform verifier feature on. Bummer, will pause on that.

Hopefully we'll get that verifier release out soon :crossed_fingers: 

If you wanted to get a head start you could add a cargo patch on this branch that points the rustls-platform-verifier dep to Ralith's branch. We can work through the other aspects of the update and then pause until that patch can be removed.

> For provider, my understanding is this is (implicitly) relying on CryptoProvider::get_default_or_install_from_crate_features() for the examples/tests

I think some of the other unit tests (like the builder ones) are going to panic with a "no process-level CryptoProvider available -- call CryptoProvider::install_default() before this point" message unless adjusted to do something similar.

cpu: > Yep just figured that out when I tested with the platform verifier feature on. Bummer, will pause on that.

@howardjohn a [compatible rustls-platform-verifier](https://github.com/rustls/rustls-platform-verifier/releases/tag/v%2F0.3.0) release is now available.

howardjohn: Ok bumped it. I made a bunch of changes, not sure they all make sense:
* Default to aws-lc-rs to match the other libraries
* Tweak tests we run in CI. --all-features and --no-default-features do not make sense since they pick 2 or 0 providers; we want 1? This aligns with tokio-rustls
* Change the `with_native_roots` and similar to not require `ring`

LMK if this makes sense

cpu: I started to look at this independently (sorry to duplicate effort!) and found it took a bit of care overall. I ended up with a branch that had a tidier commit history so I've opened a possible replacement PR: https://github.com/rustls/hyper-rustls/pull/266

cpu: > --no-default-features do not make sense since they pick 2 or 0 providers; we want 1?

Just to note here: I think we do want `--no-default-features` to work so that you can use the `with_provider_xxx` fns to bring your own `CryptProvider` without needing to build w/ aws-lc-rs or ring.

howardjohn: > Just to note here: I think we do want --no-default-features to work so that you can use the with_provider_xxx fns to bring your own CryptProvider without needing to build w/ aws-lc-rs or ring.

Good point.

Your change LGTM so happy to go with that one. Thanks for the help!

cpu: np! Thanks again for getting the ball rolling (and taking a look at my diff).

cpu:
Closing in favour of https://github.com/rustls/hyper-rustls/pull/266 - thanks again!

PR #266, opened 2024-03-25
0.27.0 preparation

cpu:
* Updates Rustls and associated ecosystem crates to use Rustls 0.23
* Fixes implicit usage of the default crypto provider
* Fixes some config gates that were overly specific to `ring` when `aws_lc_rs` would be acceptable as well
* Adds some additional CI coverage for an aws-lc-rs build, and then flips it to be the ring coverage once the default changes
* Adds the same `aws_lc_rs`/`aws-lc-rs` aliasing as used upstream
* Switches the default crypto provider to aws-lc-rs to match upstream
* Bumps the version

This is an alternative to https://github.com/rustls/hyper-rustls/pull/263 with a green build and a tidier commit history.

## Proposed release notes

* TODO - will write shortly


Cargo.toml
```diff
@@ -34,7 +34,8 @@ tokio = { version = "1.0", features = ["io-std", "macros", "net", "rt-multi-thre
 
 [features]
 default = ["native-tokio", "http1", "tls12", "logging", "ring"]
-aws-lc-rs = ["rustls/aws_lc_rs"]
+aws_lc_rs = ["rustls/aws_lc_rs"]
```
djc, line 37: Given that (a) this crate was already using the dash-based name and (b) this is a semver-incompatible release, I feel like we should either (1) do the setup the other way around and make `aws_lc_rs` the alias or (2) skip the alias and just require the hyphen-based feature name.

(Perhaps this depends on the question if we want to get rid of the underscore-based feature name in rustls 0.24.)

Cargo.toml
```diff
@@ -17,18 +17,18 @@ hyper-util = { version = "0.1", default-features = false, features = ["client-le
 log = { version = "0.4.4", optional = true }
 pki-types = { package = "rustls-pki-types", version = "1" }
 rustls-native-certs = { version = "0.7", optional = true }
-rustls-platform-verifier = { version = "0.2", optional = true }
-rustls = { version = "0.22", default-features = false }
+rustls-platform-verifier = { version = "0.3", optional = true }
```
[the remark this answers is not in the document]
cpu: Separated things so that each commit will pass.

Cargo.toml
cpu: I removed the alias in favour of requiring the hyphen-based feature name for now.

Cargo.toml
```diff
@@ -33,7 +33,7 @@ rustls-pemfile = "2"
 tokio = { version = "1.0", features = ["io-std", "macros", "net", "rt-multi-thread"] }
 
 [features]
-default = ["native-tokio", "http1", "tls12", "logging", "ring"]
+default = ["native-tokio", "http1", "tls12", "logging", "aws_lc_rs"]
```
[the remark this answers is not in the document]
cpu: Collapsed the default change into the 0.22 update commit.

src/connector/builder.rs
```diff
@@ -312,10 +321,18 @@ impl ConnectorBuilder<WantsProtocols3> {
 
 #[cfg(test)]
 mod tests {
+    fn ensure_global_state() {
+        #[cfg(feature = "ring")]
```
djc, line 325: This should go below the tests.

djc (approved):
Nice!

src/connector/builder.rs
cpu: Oops, agreed. Fixed.

ctz (approved):

cpu:
I have to run to an appointment but will do the post-merge release process when I get back in a couple of hours.

Issue #267, opened 2024-03-27
Expose feature flag to enable FIPS compliant build of AWS-LC.


tobz:
It looks like the 0.27.0 release pulled in the necessary upstream crate bumps so that AWS-LC can be used as the `rustls` backend, and with a high enough version to also use the FIPS-compliant build... 🎉 . However, users still need to do the add-your-own-dependency-to-set-the-feature-flag-you-want hack if they're using `hyper-rustls` and want to use that FIPS-compliant mode.

I plan to whip up a quick PR to expose that, but wanted to file the issue so I don't forget. 😂

djc:
Thanks for reporting -- happy to review a PR for this!

PR #268, opened 2024-03-28
Add feature flag for enabling FIPS.

tobz:
As described.

Adds a new feature flag, `fips`, to enable FIPS in `rustls`.

Verified that when enabling the new `fips` feature flag, that FIPS mode is used in `aws-lc-rs` (by seeing `aws-lc-fips-sys` being used.).

Resolves https://github.com/rustls/hyper-rustls/issues/267

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
       - id: collect-non-fips-features
         run: |
           yq --no-colors -oy '.features | keys | . - ["default", "fips"] | join(",") | "non-fips-features=" + @sh'  Cargo.toml >> "$GITHUB_OUTPUT"
-          yq --no-colors -oy '.features["default"] + ["ring"] | join(",") | "defaults-with-ring-features=" + @sh' Cargo.toml >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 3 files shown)


cpu (approved):
Thanks!

djc:
So the `fips` flag only works on Linux -- this will need some adjustments in the CI workflow to make sure we don't run `--all-features` on other platforms.

tobz:
I can try and address that here, as well. Probably also worth adding a note to the documentation that it is only supported on Linux, as well.

tobz:
Alright, pushed an update to the README. LEt me know what you think about that.

Also pushed an update to the CI to try and automatically collect the set of features to test. I did this for all-features-but-FIPS and for the "defaults + ring" features. It's definitely a little more complex, but it seems, perhaps, somewhat better from an automation standpoint so that updates to the default features don't get stale in the CI definitions.

Totally understand if it feels too complex, though: I'm happy to switch it to just be hard-coded.

tobz:
I'll just approach it that way then.

Also means less faffing about debugging the workflow since it means you having to approve them to run every single time I make a change. 😅

djc:
Yeah, using `yq` in the workflow definition is a little too meta for me, and relying on tools that aren't all that familiar to the maintainers (at least to me).

tobz:
Alright, this should be all set now.

cpu (approved):
LGTM. Thank you. The README update is a nice touch :+1:

djc (approved):

djc:
Thanks! (Bypassed the merge queue in order to squash.)

PR #269, opened 2024-04-07
Support custom server name resolution

### LINKED ISSUE #195: More elaborate custom server name
```rust
let https = HttpsConnectorBuilder::new()
            .with_native_roots()
            .https_only()
            .with_server_name("override_server_name")
            .enable_http1()
            .build();

// change it to

let https = HttpsConnectorBuilder::new()
            .with_native_roots()
            .https_only()
            .with_server_name(|request| {
                // Compute server name
                return "override_server_name";
            })
            .enable_http1()
            .build();
```

This allows different server names to be injected for different ip requests

sfackler:
I don't love the duplication of the ipv6 handling. We could have `resolve` just return a String and do the conversion in the connector maybe?

Closes #195


src/connector.rs
```diff
@@ -24,7 +25,7 @@ pub struct HttpsConnector<T> {
     force_https: bool,
     http: T,
     tls_config: Arc<rustls::ClientConfig>,
-    override_server_name: Option<String>,
+    server_name_resolver: Option<Arc<dyn ResolveServerName + Sync + Send>>,
```
djc, line 28: Not sure if it still makes sense to wrap this in `Option`? I think we should just have a `FromUriResolver` that is used by default?

I think that could also address the duplication of the IPv6 stripping.

src/connector/builder.rs
```diff
@@ -237,6 +240,22 @@ impl ConnectorBuilder<WantsProtocols1> {
...
 
+    /// Override server name for the TLS stack
+    ///
+    /// By default, for each connection hyper-rustls will extract host portion
+    /// of the destination URL and verify that server certificate contains
+    /// this value.
+    ///
+    /// If this method is called, hyper-rustls will instead use this resolver
+    /// to compute the value used to verify the server certificate.
+    pub fn with_server_name_resolver<T>(mut self, resolver: T) -> Self
```
djc, line 251: Nit: let's spell this as `impl ResolveServerName + Send + Sync + 'static`, removing the intermediate type argument name (and I think the order of traits is more conventional this way).

src/server_name_resolver.rs
```diff
@@ -0,0 +1,44 @@
+use std::error::Error;
+
+use http::Uri;
+use pki_types::ServerName;
+
+/// A trait implemented by types that can resolve a [`ServerName`] for a request.
```
djc, line 6: Since the `connector` module is pretty short anyway, I'd prefer to stick all of this stuff in the bottom of that module. Note that we prefer a top-down style, so please order as such:

```rust
1. `struct FixedServerNameResolver` + impls
2. `impl<F, E> ResolveServerName for F`
3. `pub trait ResolveServerName`
```

src/server_name_resolver.rs
```diff
@@ -0,0 +1,44 @@
...
+    F: Fn(&Uri) -> Result<ServerName<'static>, E>,
+    E: Into<Box<dyn Error + Sync + Send>>,
+{
+    fn resolve(&self, uri: &Uri) -> Result<ServerName<'static>, Box<dyn Error + Sync + Send>> {
+        self(uri).map_err(Into::into)
+    }
+}
+
+pub(crate) struct FixedServerNameResolver {
+    name: String,
```
djc, line 23: Ideally we'd hold a `ServerName` in here so that we don't have to do the conversion on every connection attempt, but I guess that's tricky to fit in a semver-compatible API. I suppose we can fix that with the next semver-compatible version.

Maybe we should just `#[deprecate]` `with_server_name()` (and make this `pub`)?

djc (commented):
This is mostly looking good, here's a bunch of suggestions.

What's your use case for this?

sfackler:
My specific use case is interacting with a [headless service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) in Kubernetes. In a headless service, the cluster publishes a DNS record mapping e.g. `my-svc.my-namespace.svc.cluster-domain.example` to the IP addresses for each replica of the service. A client that wants to talk to the service resolves the DNS record to get the set of replica IPs, and selects the one it wants to send the request to (it's running its own load balancing logic internally). However, we can't just send the request to `https://10.0.1.123/my-endpoint` since the TLS cert is only issued for `my-svc.my-namespace.svc.cluster-domain.example` and not the individual pod IPs.

To make this work, cleanest way seems to be to rewrite the URI to include both the hostname and IP. I can then create a custom DNS resolver for the Hyper client which pull the IP out of the URI, and a custom servername resolver in hyper-tls that pulls the hostname out.

src/server_name_resolver.rs
sfackler: Yep, I'll update this to take in a `ServerName` directly, and then deprecate with_server_name and update it to just use a 1-off closure.

src/connector.rs
sfackler: Sure, will add a default resolver.

src/connector.rs
```diff
@@ -147,3 +133,76 @@ impl<T> fmt::Debug for HttpsConnector<T> {
             .finish()
     }
 }
+
+/// The default server name resolver, which uses the hostname in the URI.
+#[derive(Default)]
+pub struct DefaultServerNameResolver(());
+
+impl DefaultServerNameResolver {
```
djc, line 141: Nit: I would prefer ditching this `impl` and relying on `Default` only.

djc (approved):
Nice!

src/connector.rs
sfackler: removed!

src/connector/builder.rs
```diff
@@ -237,6 +242,22 @@ impl ConnectorBuilder<WantsProtocols1> {
...
     }
 
+    /// Override server name for the TLS stack
+    ///
+    /// By default, for each connection hyper-rustls will extract host portion
+    /// of the destination URL and verify that server certificate contains
+    /// this value.
+    ///
+    /// If this method is called, hyper-rustls will instead use this resolver
+    /// to compute the value used to verify the server certificate.
```
cpu, line 252: For consistency with the `with_server_name` docs it feels worth keeping the explicit mention of the destination URL being ignored.

```suggestion
    /// to compute the value used to verify the server certificate. Domain name included in
    /// the URL will not affect certificate validation.
```

src/connector/builder.rs
```diff
@@ -246,9 +267,19 @@ impl ConnectorBuilder<WantsProtocols1> {
     /// If this method is called, hyper-rustls will instead verify that server
     /// certificate contains `override_server_name`. Domain name included in
     /// the URL will not affect certificate validation.
-    pub fn with_server_name(mut self, override_server_name: String) -> Self {
-        self.0.override_server_name = Some(override_server_name);
-        self
+    #[deprecated(since = "0.27.1", note = "use Self::with_server_name_resolver instead")]
```
cpu, line 270: I think it'd be helpful to mention the fixed name resolver here since that's the way to achieve the replacement of `with_server_name` with minimal changes:

```suggestion
    #[deprecated(since = "0.27.1", note = "use Self::with_server_name_resolver with FixedServerNameResolver instead")]
```

cpu (approved):
Thanks! Looks good to me. I had a couple of non-blocking suggestions to consider.

src/connector/builder.rs
sfackler: This case is a bit difference since the URL's domain name may affect certificate validation depending on how the resolver behaves, right?

cpu: That's a fair point. Leaving it as is sounds good to me :+1:

cpu (approved):

src/connector/builder.rs
cpu: > rustls / Format (pull_request) Failing after 17s

Sorry, looks like my suggested fix needs to go through `rustfmt` to appease CI.

sfackler: fixed!

cpu:
@djc should we cut a release for this?

djc:
I was defaulting to not doing so (proactively) unless the OP asks for it.

PR #273, opened 2024-04-26
Avoid `aws-lc-rs` feature flag when building docs.

### LINKED ISSUE #272: Latest version (0.27.1) fails to build for `docs.rs`
As seen here: https://docs.rs/crate/hyper-rustls/0.27.1/builds/1185949

This looks to be directly related to my changes in #268, and I'm going to try and take a look at it today.

tobz:
As stated in the title.

Mimics the fix in rustls/rustls#1827 by simply enabling all features manually, minus `aws-lc-rs`. This avoids trying to build it at in during docs.rs generation, as docs.rs _can't_ currently compile it, whether or not FIPS is enabled.

There's nothing in the API, though, that depends on the feature flag so we didn't have to do anything in the code itself to cope with not passing the `aws-lc-rs` feature flag when building docs.

Tested locally using the docs.rs [build instructions](https://github.com/rust-lang/docs.rs/blob/master/README.md#build-subcommand) and verified that the build was broken prior to this fix, and the build completed successfully with this fix.

Fixes #272.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,5 +54,6 @@ path = "examples/server.rs"
 required-features = ["aws-lc-rs"]
 
 [package.metadata.docs.rs]
-all-features = true
+no-default-features = true
+features = ["http1","http2","webpki-tokio","native-tokio","ring","tls12","logging"]
 rustdoc-args = ["--cfg", "docsrs"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,5 +55,5 @@ required-features = ["aws-lc-rs"]
 
 [package.metadata.docs.rs]
 no-default-features = true
-features = ["http1","http2","webpki-tokio","native-tokio","ring","tls12","logging"]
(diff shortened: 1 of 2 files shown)


cpu (approved):
Thank you

Cargo.toml
```diff
@@ -54,5 +54,6 @@ path = "examples/server.rs"
 required-features = ["aws-lc-rs"]
 
 [package.metadata.docs.rs]
-all-features = true
+no-default-features = true
+features = ["http1","http2","webpki-tokio","native-tokio","ring","tls12","logging"]
```
djc, line 58: Tiny nit: can you just add spaces after the commas, as is conventional? (Unlike in command-line multi-arguments.)

tobz: Addressed in 0a35039.

djc:
Bypassed queue in order to sqaush during merge.

PR #279, opened 2024-07-11
enable `rustls-platform-verifier` for docs.rs

flisky:


cpu (approved):
Good idea, thank you!

djc (approved):

djc (commented):
Thanks -- would you mind keeping them in alphabetic order?

In rustls we've made the jump to `#![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]`, that would be even better to follow here, if you want to spend a little time on that.

flisky:
It seems `doc_auto_cfg` is already there in f9775bcc1c30b9a0c620ea6c623bbd8025e8494b by yourself, @djc:)

djc:
Okay, thanks for the sorting!

PR #283, opened 2024-08-28
Builder fixes

### LINKED ISSUE #282: make `with_provider_and_*` take `Arc<CryptoProvider>` instead of `CryptoProvider`
Suggested by `Adrian` in [discord](https://discord.com/channels/976380008299917365/1015156984007381033/1272237736132087871):
> Hey there, seems like hyper-rustls' with_provider_and_* methods take a CryptoProvider, meaning that it's impossible to call CryptoProvider::get_default() in a library and pass that as the argument (since getting an owned value out of the Arc is not really possible). I think the API should take an Arc<CryptoProvider>, just like rustls' ClientConfig::with_provider, which it calls internally (by creating an Arc via .into()).

Ctz mentioned _impossible_ might be an overstatement (e.g. due to [`Arc::unwrap_or_clone()`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.unwrap_or_clone)) but in general there was agreement that the API should be changed to take an `Arc`.

We should consider making this change alongside future breaking API changes.

Gelbpunkt:
These are two changes to make the most recent hyper-rustls release usable for us in [twilight](https://github.com/twilight-rs/twilight). We also ran into the issue that was fixed by commit https://github.com/rustls/hyper-rustls/commit/c56c76a99ab0e9413fba5cc04df2d4fcbfd45735 in our CI, I would be very grateful for a new release with these commits to allow us to finally upgrade rustls.

The first commit adds a method that I think was overlooked when the rustls-platform-verifier support was initially added.

The second commit resolves #282 in a backwards-compatible manner.

### DIFF
--- a/src/connector/builder.rs
+++ b/src/connector/builder.rs
@@ -1,7 +1,11 @@
 use std::sync::Arc;
 
 use hyper_util::client::legacy::connect::HttpConnector;
-#[cfg(any(feature = "rustls-native-certs", feature = "webpki-roots"))]
+#[cfg(any(
+    feature = "rustls-native-certs",
+    feature = "rustls-platform-verifier",
+    feature = "webpki-roots"
+))]
 use rustls::crypto::CryptoProvider;
 use rustls::ClientConfig;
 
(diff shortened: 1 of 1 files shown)


djc (approved):
LGTM, thanks!

cpu (approved):

PR #285, opened 2024-08-30
Upgrade to rustls-native-certs 0.8

djc:


ctz (approved):

src/config.rs
```diff
@@ -47,13 +51,24 @@ impl ConfigBuilderExt for ConfigBuilder<ClientConfig, WantsVerifier> {
...
-        {
+        let CertificateResult { certs, errors, .. } = rustls_native_certs::load_native_certs();
+        if !errors.is_empty() {
+            crate::log::warn!("native root CA certificate loading errors: {errors:?}");
+        }
+
+        if certs.is_empty() {
+            return Err(io::Error::new(
+                io::ErrorKind::NotFound,
+                "no native root CA certificates found (errors: {errors})",
```
nickelc, line 67:
```suggestion
                format!("no native root CA certificates found (errors: {errors})"),
```

djc:
"Simplify log macro setup" tries to make the log shim in this crate more like [rustls'](https://github.com/rustls/rustls/blob/main/rustls/src/lib.rs#L383), but so far I can't quite get it to work. As reported by CI, `cargo test --no-default-features --features rustls-native-certs` fails with

```rust
error[E0433]: failed to resolve: could not find `warn` in `log`
  --> src/config.rs:61:25
   |
61 |             crate::log::warn!("native root CA certificate loading errors: {errors:?}");
   |                         ^^^^ could not find `warn` in `log`
```

(Not sure what's different about these crates? What am I missing?)

If I skip that commit, I get this error:

```rust
error[E0659]: `warn` is ambiguous
  --> src/lib.rs:59:20
   |
59 |     pub(crate) use warn;
   |                    ^^^^ ambiguous name
   |
   = note: ambiguous because of a name conflict with a builtin attribute
   = note: `warn` could refer to a built-in attribute
note: `warn` could also refer to the macro defined here
  --> src/lib.rs:57:5
   |
57 |     macro_rules! warn    ( ($($tt:tt)*) => {{}} );
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

If someone else wants to take a look, that would be great!

ctz:
> (Not sure what's different about these crates? What am I missing?)

I never got this to work as desired in rustls, though I think I now understand more about how macro visibility works. Over there it works in a weird way:

- with feature "logging", code should import `crate::log::warn` and then call `warn!("oh no");`
- without feature "logging", code should _not_ import anything, but still can call `warn!("oh no");`

in the first case, the `warn` macro is imported from the log crate. in the second, it relies on `#[macro_use]` globalising the macro name within the crate (which I guess also works around the collision with the warn attribute.)

pushed a commit which avoids all that by keeping the macros local, and explicitly `use`ing them to define their visibility, and then avoiding `warn` as a macro name.

djc:
Awesome, thanks! I took care of the squashing.

src/config.rs
djc: Thanks!

cpu (approved):
What happened with ef0f5eac69d85cb818aee59e35e05c3f5430b079 ? It renders empty both in GitHub UI and when I `git diff`/`git show` it locally :thinking:

djc:
> What happened with [ef0f5ea](https://github.com/rustls/hyper-rustls/commit/ef0f5eac69d85cb818aee59e35e05c3f5430b079) ? It renders empty both in GitHub UI and when I `git diff`/`git show` it locally 🤔

Oops, took that commit out.

ctz (approved):

djc:
* Published hyper-rustls v0.27.3 at registry `crates-io`
* [new tag]         v/0.27.3 -> v/0.27.3
* [Release notes](https://github.com/rustls/hyper-rustls/releases/tag/v%2F0.27.3)

Didn't call anything out in the release notes since it's just a grab bag of smaller stuff.

PR #289, opened 2024-09-20
perf: forward write-vectored calls

seanmonstar:


src/stream.rs
```diff
@@ -83,6 +83,26 @@ impl<T: rt::Write + rt::Read + Unpin> rt::Write for MaybeHttpsStream<T> {
         }
     }
 
+    #[inline]
+    fn poll_write_vectored(
```
cpu, line 87: nit: Can we keep the fn order in the impl matching the order from the `rt::Write` trait? e.g. poll_write, poll_flush, poll_shutdown, and then the two new fns: is_write_vectored followed by `poll_write_vectored`.

cpu (approved):
Thanks :+1:

src/stream.rs
seanmonstar: Done!

cpu: Thanks!

djc (approved):

PR #293, opened 2024-12-17
Actually use configured crypto provider for the platform verifier

Gelbpunkt:
`rustls-platform-verifier`'s `Verifier` needs to be told about the provider that is being used, otherwise it would error if no process-wide default had been installed. `with_provider_and_platform_verifier` does create the `ClientConfig` with the crypto provider and then calls `with_platform_verifier`, but the platform verifier was previously never made aware of the crypto provider.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hyper-rustls"
-version = "0.27.3"
+version = "0.27.4"
 edition = "2021"
 rust-version = "1.71"
 license = "Apache-2.0 OR ISC OR MIT"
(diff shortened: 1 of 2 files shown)


djc (approved):
Ohh, thanks for that! Silly mistake...

Can you add in a commit that bumps the version number?

cpu (approved):

Gelbpunkt:
> Can you add in a commit that bumps the version number?

Done :)

ctz (approved):

djc:
* Published hyper-rustls v0.27.4 at registry `crates-io`
* [new tag]         v/0.27.4 -> v/0.27.4
* [Release notes](https://github.com/rustls/hyper-rustls/releases/tag/v%2F0.27.4)

PR #296, opened 2024-12-20
Prepare 0.27.5 release

djc:
## Proposed release notes

Refactoring in #245 (first released in 0.25.0) broke the optional enforcement of HTTPS only connections (that is, only HTTPS connections can be made through the connector) on the hyper-rustls `HttpsConnector`; this was found and fixed in #295.

Follow-up from #295.

### DIFF
--- a/src/connector.rs
+++ b/src/connector.rs
@@ -220,6 +220,11 @@ mod tests {
     use crate::{ConfigBuilderExt, HttpsConnectorBuilder};
 
     fn tls_config() -> rustls::ClientConfig {
+        #[cfg(feature = "rustls-platform-verifier")]
+        return rustls::ClientConfig::builder()
+            .with_platform_verifier()
+            .with_no_client_auth();
+
         #[cfg(feature = "rustls-native-certs")]
         return rustls::ClientConfig::builder()
             .with_native_roots()
@@ -229,11 +234,6 @@ mod tests {
         #[cfg(feature = "webpki-roots")]
         return rustls::ClientConfig::builder()
             .with_webpki_roots()
-            .with_no_client_auth();
-
-        #[cfg(feature = "rustls-platform-verifier")]
-        return rustls::ClientConfig::builder()
-            .with_platform_verifier()
             .with_no_client_auth();
     }
 
--- a/src/connector.rs
+++ b/src/connector.rs
@@ -251,29 +251,23 @@ mod tests {
         assert_eq!(message, "unsupported scheme http");
     }
 
-    fn tls_config() -> rustls::ClientConfig {
-        #[cfg(feature = "rustls-platform-verifier")]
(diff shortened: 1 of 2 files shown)


src/connector.rs
```diff
@@ -245,29 +245,29 @@ mod tests {
...
-            .with_webpki_roots()
-            .with_no_client_auth();
-    }
-
     async fn connect(
         https_only: bool,
         https: bool,
     ) -> Result<MaybeHttpsStream<TokioIo<TcpStream>>, BoxError> {
-        let builder = HttpsConnectorBuilder::new().with_tls_config(tls_config());
+        let config_builder = rustls::ClientConfig::builder();
```
ctz, line 252: consider cfg_if? unused code warnings aside, this is a lot less readable than before, to my eye

ctz (approved):

src/connector.rs
djc: Good idea -- how about now?

ctz (approved):

src/connector.rs
```diff
@@ -221,28 +221,28 @@ mod tests {
 
     #[tokio::test]
     async fn connects_https() {
-        oneshot(https_or_http_connector(), https_uri())
+        oneshot(https_or_http_connector(), true)
```
cpu: Maybe a two variant enum instead of a `bool`? I think the previous version with the helper made it more obvious which invocation was being handed a `https://` uri vs an `http://` uri.

src/connector.rs
```diff
@@ -284,10 +263,18 @@ mod tests {
             .with_no_client_auth();
     }
 
-    async fn oneshot<S: Service<Uri>>(
-        mut service: S,
+    async fn connect(
+        https_only: bool,
         https: bool,
```
cpu: similar nit here: two `bool`'s side-by-side is easy to get wrong at the call-site.

cpu (approved):
Thanks :+1: 

~~Can you propose some release notes as well?~~ Oops; they're there. Sorry! Notes look good to me.

src/connector.rs
djc: I considered it, but it feels a little like overkill? This is not library API but just a test function with the tests right next to it? I doubt these functions will see much change...

cpu: I don't see the harm in being more explicit, especially in an area where there was a previous oversight. It's a minor nit though, so whatever you want is fine with me.

djc: Never mind, I think the enums are worth it.

cpu (approved):

djc:
* Published hyper-rustls v0.27.5 at registry `crates-io`
* [new tag]         v/0.27.5 -> v/0.27.5
* [Release notes](https://github.com/rustls/hyper-rustls/releases/tag/v%2F0.27.5)

PR #301, opened 2025-05-31
Add From<(H, C, Arc<dyn ResolveServerName + Send + Sync>)> to HttpsConnector

MarkusTieger:
As you can see here ,

HttpsConnector implements
```
From<(H, C)>  for HttpsConnector<H>
where
    C: Into<Arc<rustls::ClientConfig>>
```
which just sets the server_name_resolver to `DefaultServerNameResolver`.

This adds
```
impl<H, C> From<(H, C, Arc<dyn ResolveServerName + Send + Sync>)> for HttpsConnector<H>
where
    C: Into<Arc<rustls::ClientConfig>>
```
which does the same, but can be used to specify another server_name_resolver.


cpu (approved):
Seems pretty reasonable to me. Thanks!

djc (commented):
Hmm, I'm inclined to add a method for this to the builder instead of adding yet another `From` impl. Tuples get a bit unwieldy and this isn't quite a full solution since it doesn't support `force_https: true`, for example.

djc:
Actually, it looks like you can already do this? The `ConnectorBuilder` in `WantsProtocols1` state has a `with_server_name_resolver()` method.

MarkusTieger:
Yeah, but the builder doesn't accept a TlsConfig with a predefined alpn_protocols. Thats, at least to me, the main use of the From implementation that exists already.

djc:
Just for my understanding, why do you need/want to set `alpn_protocols` yourself?

MarkusTieger:
Also, does the extra From implementation in the code base really hurt that much?

djc:
> Also, does the extra From implementation in the codebase really hurt that much?

It does not hurt that much -- but it's not great API in the first place and we're potentially stuck with it for a long time, so I'd rather be careful about adding public API. And given that it's not yet clear how reqwest might accomodate your use case, we probably want for that to become more clear.

djc:
Yeah, I'd be more comfortable with that (which I'm aware might be silly).

cpu (approved):
If djc likes this way better I'm happy with it too :-) Thanks!

MarkusTieger:
Is there a blocker? This pr has been unmerged and open for 3 months ...

ctz (approved):

ctz:
Sorry -- my fault.

PR #303, opened 2025-06-05
Upgrade to rustls-platform-verifier 0.6

djc:
## Proposed release notes

- Seal the `ConfigBuilderExt` trait. This is an extension trait used to offer a more convenient server verifier configuration API. This is technically a breaking change, but we think it is unlikely that anyone has actually implemented this trait.
- Upgrade to rustls-platform-verifier 0.6. Because the platform verifier is now initializing its root certificate store on some platforms eagerly (on initialization rather than on first use), infallible API for setting up the platform verifier has been deprecated in favor of newly added fallible API.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hyper-rustls"
-version = "0.27.6"
+version = "0.27.7"
 edition = "2021"
 rust-version = "1.71"
 license = "Apache-2.0 OR ISC OR MIT"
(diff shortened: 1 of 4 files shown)


djc:
> Do we want to seal `ConfigBuilderExt`? I think that would make this a non-breaking change.

In the future, yes -- sealing it would be breaking now.

ctz (approved):

djc:
* Published hyper-rustls v0.27.7 at registry `crates-io`
* [new tag]         v/0.27.7 -> v/0.27.7
* [Release notes](https://github.com/rustls/hyper-rustls/releases/tag/v%2F0.27.7)

PR #357, opened 2026-06-09
Port hyper-rustls to rustls 0.24

iadev09:
This ports hyper-rustls to the current rustls 0.24 API

ref:  quinn-rs/quinn#2671


djc:
Thanks!

I think we'll want to hold off on merging this for now.

